PR #3, opened 2022-07-13
Refactor Copilot suggestions to provide better completion experience

timfjord:
The original idea of this PR was to slightly improve Copilot's suggestion and remove the part of the suggestion that is already present in the line. But when I started working on it I realised that it would require some refactoring as the package wasn't designed to handle suggestions in that way. So I ended up with this PR. 
Знімок екрана 2022-07-13 о 13 47 10

Here is a short summary of what this PR does:

- parses suggestions and removes the part of the text that it is already visible(this is only the case for completion inside line). This provides a much better user experience because in most cases you can use the suggestion as it is as previously you had to correct all accepted suggestions from the middle of the line 
- keeps the current suggestion in the view settings that allows having a custom context and binding `Tab`/`Esc` to accept/dismiss the current suggestion
- introduces the `ui.Completion` class with all completion-related logic (including `PopupCompletion` with popup logic)
- minor styling improvements (some margins, etc)
- reorganises the structure a bit(mostly to support new changes)
- default key bindings(`Tab` to accept, `Esc` to dismiss). We could comment them by default but keep them in the file so users can copy and paste them if needed

There are still some edge cases. For example, sometimes Copilot suggests a completely different end of the line. I will keep an eye on such cases and create a separate PR as it might require some time to collect all of them and compare how VSCode/neovim handles them.

### DIFF
--- a/commands.py
+++ b/commands.py
@@ -1,23 +1,15 @@
-from .constants import (
-    PACKAGE_NAME,
-    REQ_CHECK_STATUS,
-    REQ_SIGN_IN_CONFIRM,
-    REQ_SIGN_IN_INITIATE,
-    REQ_SIGN_OUT,
-)
-from .plugin import CopilotPlugin
-from .types import (
-    CopilotPayloadSignInConfirm,
-    CopilotPayloadSignInInitiate,
-    CopilotPayloadSignOut,
-)
-from .ui import Completion
 from abc import ABCMeta
+
+import sublime
+import sublime_plugin
 from LSP.plugin import Request
 from LSP.plugin.core.registry import LspTextCommand
 from LSP.plugin.core.typing import Tuple, Union
-import sublime
-import sublime_plugin
+
+from .constants import PACKAGE_NAME, REQ_CHECK_STATUS, REQ_SIGN_IN_CONFIRM, REQ_SIGN_IN_INITIATE, REQ_SIGN_OUT
+from .plugin import CopilotPlugin
+from .types import CopilotPayloadSignInConfirm, CopilotPayloadSignInInitiate, CopilotPayloadSignOut
+from .ui import Completion
 
 
 class CopilotInsertAsIsCommand(sublime_plugin.TextCommand):
--- a/listeners.py
+++ b/listeners.py
@@ -1,14 +1,15 @@
+import functools
+
 import sublime
+import sublime_plugin
+from LSP.plugin.core.types import FEATURES_TIMEOUT, debounced
+
 from .plugin import CopilotPlugin
 from .ui import Completion
-from LSP.plugin.core.types import debounced
-from LSP.plugin.core.types import FEATURES_TIMEOUT
-import functools
-import sublime_plugin
 
 
 class EventListener(sublime_plugin.ViewEventListener):
-    COPILOT_SUGGESTION_VISIBLE = 'copilot_suggestion_visible'
+    COPILOT_SUGGESTION_VISIBLE = "copilot_suggestion_visible"
 
     def on_modified_async(self) -> None:
         plugin = CopilotPlugin.plugin_from_view(self.view)
--- a/plugin.py
+++ b/plugin.py
@@ -1,3 +1,12 @@
+import functools
+import os
+import weakref
+
+import sublime
+from LSP.plugin import Request, Session, filename_to_uri
+from LSP.plugin.core.typing import Optional, Tuple
+from lsp_utils import ApiWrapperInterface, NpmClientHandler, notification_handler
+
 from .constants import (
     COPILOT_WAITING_COMPLETION_KEY,
     NTFY_LOG_MESSAGE,
@@ -16,17 +25,6 @@ from .types import (
 )
 from .ui import Completion
 from .utils import get_project_relative_path
-from LSP.plugin import filename_to_uri
-from LSP.plugin import Request
-from LSP.plugin import Session
-from LSP.plugin.core.typing import Optional, Tuple
-from lsp_utils import ApiWrapperInterface
-from lsp_utils import notification_handler
-from lsp_utils import NpmClientHandler
-import functools
-import os
-import sublime
-import weakref
 
 
 def plugin_loaded():
@@ -180,6 +178,4 @@ class CopilotPlugin(NpmClientHandler):
         if not completions:
             return
 
-        sublime.set_timeout_async(
-            lambda: Completion(view).show(region, completions)
-        )
+        sublime.set_timeout_async(lambda: Completion(view).show(region, completions))
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,6 +13,7 @@ ignore_missing_imports = true
 
 [tool.isort]
 profile = 'black'
+line_length = 120
 
 [tool.black]
 line-length = 120
--- a/ui.py
+++ b/ui.py
@@ -1,40 +1,40 @@
 import textwrap
-from .constants import COPILOT_VIEW_SETTINGS_PREFIX
-from .types import (
-    CopilotPayloadCompletion,
-)
-from LSP.plugin.core.typing import List, Tuple
-import sublime
+
 import mdpopups
+import sublime
+from LSP.plugin.core.typing import List, Tuple
+
+from .constants import COPILOT_VIEW_SETTINGS_PREFIX
+from .types import CopilotPayloadCompletion
 
 
 class Completion:
     def __init__(self, view: sublime.View):
         self.view = view
 
-    def _settings(self, key, default_or_new_value=None, do: str = 'get'):
-        settings_key = '{}.{}'.format(COPILOT_VIEW_SETTINGS_PREFIX, key)
+    def _settings(self, key, default_or_new_value=None, do: str = "get"):
+        settings_key = "{}.{}".format(COPILOT_VIEW_SETTINGS_PREFIX, key)
 
-        if do == 'set':
+        if do == "set":
             return self.view.settings().set(settings_key, default_or_new_value)
-        elif do == 'erase':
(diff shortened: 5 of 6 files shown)


TheSecEng:
Do we need to store the current completion in view settings ? I don't believe so, we should be able to get a key binding with context to be targeted for popup. 


Another thought is once we get the context working, we should implement a prev and next suggestion sincecopilot returns a list of suggestions and we need to support cycling. This is a separate PR though

timfjord:
We need to parse the text before dispalying it.
Say, on the screenshot above, in the original implementation it would have been `set')` instead of just `set'` and if it would be accepted we would make the line invalid, so we cannot rely on the raw suggestion anymore that's why I keep the "real" display text in the view settings.

As for the prev/next, I thought about that too. By idea was to keep the current cycle(index) and the list of completions in settings too, and then `Completion` `next`/`previous classes` to cycle thought them. 
Theoretically, it should be as simple as(of course if we make sure we keep completions and the current index in the settings):
```python
def next(self):
    region = self.region
    completions = self.completions
    
    return self.show(region, completions, self.cycle + 1)
```

Also, I am planning to work on inline suggestions(like in VSCode, neovim) I am almost sure it should be possible, but maybe with some minor limitations, so I would rather rely on more abstract things when determining context

timfjord:
Also, regarding this line

> There are still some edge cases. For example, sometimes Copilot suggests a completely different end of the line. I will keep an eye on such cases and create a separate PR as it might require some time to collect all of them and compare how VSCode/neovim handles them.

I checked how VSCode handles it and it is really cool. Basically, they split suggestions into separate regions.
It is more an edge case, so I would handle it separately, but it should be possible with our implementation too.

TheSecEng:
> We need to parse the text before dispalying it.
> 
> Say, on the screenshot above, in the original implementation it would have been `set')` instead of just `set'` and if it would be accepted we would make the line invalid, so we cannot rely on the raw suggestion anymore that's why I keep the "real" display text in the view settings.

We should still be able to store current suggestions in the Copilot class for the view and then adjust on the fly for display and completions. When a completion is triggered it would do all that logic. 

Storing these in the view settings doesn't really give a warm and fuzzy. 

I'm sure @jfcherng probably has a better explanation

jfcherng:
> Say, on the screenshot above, in the original implementation it would have been `set')` instead of just `set'` and if it would be accepted we would make the line invalid, so we cannot rely on the raw suggestion anymore that's why I keep the "real" display text in the view settings.

It didn't work for this case somehow.
image
image

Probably related to auto bracket pairing when pressing `(`.

timfjord:
Yes, only Copilot files will be reloaded, so if you have something like
```python
_storage = {}

class Test:
    def get_data(self, view):
        global _storage

        if not _storage.get(view.id):
            _storage[view.id] = { ... }

        return _storage[view.id]
```
The `_storage` will be reloaded when the file is reloaded. But again, I don't think it is critical.

On the other hand, keeping data like this in `view.settings` is pretty common. I've seen this in other packages, but maybe there are some limitations that I am not aware of.

As far as I know, this `settings` thing is never persisted in the file system. It is basically a dictionary related to the current view, with all the settings copied to it

jfcherng:
> This limitation I mentioned above is not critical, I am sure we can live with it and I am happy to adjust the PR

Yes, generally I like concepts/improvements in this PR. It's not perfect but it's better UX for sure. I think further improvements can be done in another PRs if any.

TheSecEng:
@timfjord ill do a bit of testing today and I believe this can be merged after. 

Ills create some issues to track further improvements. 


Thanks for your work in this.

timfjord:
Awesome, thanks, guys 🎉

My next priority will be inline suggestions then!

jfcherng:
I may run the `black` formatter on the whole project again for this PR. And revise things like

```diff
-if (self.view.classify(region[1]) & sublime.CLASS_LINE_END) != 0:
+if self.view.classify(region[1]) & sublime.CLASS_LINE_END:
```

timfjord:
Absolutely, I think we should add some tooling(`flake8`, `black` and definitely tests)

jfcherng:
The formatted result is a bit odd to me (personally I didn't use `isort` but just use something like F9 to do sorting things). I guess we need to setup this https://pycqa.github.io/isort/docs/configuration/black_compatibility.html#using-a-config-file-such-as-isortcfg ?

timfjord:
Hm, indeed, it uses 99 lines by default. Let me try to change that and use `black` profile

timfjord:
I incidentally squashed all commits but I think it is looking good now. Please take a look 

As far as I understand, `isort` adds all those `(` / `)` because it uses `line_length = 79`, so this is how it tries to satisfy this requirement

jfcherng:
> I incidentally squashed all commits but I think it is looking good now. Please take a look
> 
> As far as I understand, `isort` adds all those `(` / `)` because it uses `line_length = 79` by default, so this is how it tries to satisfy this requirement

Adding `()` is fine, but black won't align the next line to the opening `(`. Anyway, I added a setting to make isort compatible with black now, so 👍

jfcherng:
I guess GitHub is drunken now. https://github.com/TheSecEng/LSP-copilot/pull/3/files only show differences from your latest (force-pushed) commit but not the differences of this PR.

timfjord:
Hm, you could try to toggle `Viewed` checked. I remember after I force-pushed the changes all changes were folded and I thought there is no difference, but I after I uncheck `Viewed` I saw the diff.

timfjord:
@jfcherng Please let me know if you are still experiencing issues with the diff view.  
I think I can just push to another branch and create another PR.

Theoretically, the diff should be also viewed from here https://github.com/TheSecEng/LSP-copilot/pull/3/commits/f55c695efe3be6ddc8d215eae8485523b8969326

jfcherng:
Inserting inline completion is possible via removing the line and insert the full completion text. (https://github.com/TheSecEng/LSP-copilot/commit/48e07ab7b8320830b911dbfbdf03be1d1e15914a)

https://user-images.githubusercontent.com/6594915/179369133-2b4fca9c-f46c-4d12-ad55-1ff568be87c7.mp4

There is no decent way to show the "ghost text" when it's a inline completion though. Making me feel displaying the full text provides better UX?
image v.s. image

timfjord:
Cool, I was thinking about something similar. I haven't tested https://github.com/TheSecEng/LSP-copilot/commit/48e07ab7b8320830b911dbfbdf03be1d1e15914a but there were quite a few issues with the current approach and [your fix](https://github.com/TheSecEng/LSP-copilot/commit/48e07ab7b8320830b911dbfbdf03be1d1e15914a) should potential fix them. My idea was to split the dispaly_text and  build a set of regions with substrings, but it looks like https://github.com/TheSecEng/LSP-copilot/commit/48e07ab7b8320830b911dbfbdf03be1d1e15914a is simpler and should do the job 👍🏻 

As for displaying the `dispaly_text` vs `text`. Since we use popup I think both make sense. 
I personally even lean more towards dispalying the entire text as it might be more intuitive.  
But what if we make this configurable?  
Say, if, `suggestion.popup.dispaly_full_text` is `true` we do as on the right screenshot, otherwise as on the left(default)?

Also, last week I played a bit with phantom sets and tried to achieve the ghost text suggestions look-and-feel(I call them inline suggestion).  I even found a workaround how to insert a phantom but keep the cursor as it is(basically you need to insert a blank `LAYOUT_BLOCK` or `LAYOUT_BELOW` phantom), but there were still other issues(especially if the cursor is in the middle of the line). The implementation is far from ideal and I will try to share something testable next week.

But the main problem is not how we display but more how we work with the existing dispaly text. For example, in VSCode, new requests aren't triggered if there is an inline suggestion and you type characters that match the beginning of the suggestion(I think it is handled internally by VSCode API).  
I think we should try to mimic that, even that currently we don't do ghost text. This should also potentially fix the flickering effect where the same popup appears twice.

timfjord:
@jfcherng I've been testing this new replacement implementation and I think we should definitely should the entire line(like on the right screenshot. Currently, it can be quite confusing because we show only the difference but when we accept the completion the entire line gets replaced.

Ideally, we should calculate a set of regions and insert only the required parts of the completion(this way the cursor won't jump to the end of the line each time we accept a completion). But since it would require some time to implement, we should show the entire completion in the popup as a compromise.

jfcherng:
> Ideally, we should calculate a set of regions and insert only the required parts of the completion(this way the cursor won't jump to the end of the line each time we accept a completion).

Where should the caret be after completion? Since for an inline completion, there can be multiple candidate positions.
image

TheSecEng:
I would assume the end if the completion

timfjord:
Well, for multiline completion yes, and it is perfectly fine to move the cursor to the end of the inserted text(like we do now).

My concern was more about single line complections that we insert in the middle of the line. 
In the previous(not perfect implementation) we inserted only that part of the text that was reqruied and the curosr moved the end of inserted piece, currently we move the cursor to the end of the line, no matter what content we insert.

For example, if there something like 
```
my_method(arg1, a);
                 ^ cursor is here and the completion is arg2
```
in the previous implementation, the cursor would have been moved here
```
my_method(arg1, arg2);
                    ^
```
and in the current implementation(since we are replacing the entire line) the cursor would be here
```
my_method(arg1, arg2); 
                      ^
```

The previous implementation wasn't perfect and there were issues with it. I have ideas how to improve it, but it would require some time.
What I am suggesting is to display the popup differently like you showed here

when we show the entire block of text

jfcherng:
> What I am suggesting is to display the popup differently like you showed here
> 
> when we show the entire block of text

Yeah. That also resolves the syntax highlighting issue. It's a pity that it doesn't work well with a long line. 😢

PR #4, opened 2022-07-13
Notify Accepted / Notify Rejected

TheSecEng:
### What

- Implements the concept of `Notify Accepted` and `Notify Rejected` Copilot requests
- Blocks `Notify Accepted` and `Notify Rejected` requests behind the `telemetry` setting (_default_: false)
- Adds a `sublime-package.json` file to support setting smart completion


jfcherng:
I just resolved conflicts here due to my changes on `master`.

commands.py
```diff
@@ -18,6 +19,11 @@ def want_event(self) -> bool:
         return False
 
 
+class CopilotInsertAsIsCommand(CopilotTextCommand):
```
timfjord, line 22: I think this command is no longer required

TheSecEng:
Not currently happy with this implementation seeing as

1. Completions aren’t stored anywhere
2. We default to the first completion always
3. Passing settings into Completion class right now isn’t possible.

commands.py
jfcherng: Resolved in f5f0ebd

TheSecEng:
Yes, it is telemetry. And I am putting them behind a setting flag. Which I will push momentarily. They help improve completion accuracy

commands.py
```diff
@@ -27,17 +28,53 @@ def run(self, edit: sublime.Edit) -> None:
...
         display_text = completion.display_text or ""
+        completion_uuid = completion.uuid or ""
 
         if not region:
             return
 
         completion.hide()
         self.view.insert(edit, region[1], display_text)
 
+        session = self.session_by_name(self.session_name)
```
timfjord, line 39: The logic here is almost identical(except the Request name part), so we could extract it to a private method `_report_telemetry` and call it in both commands.

commands.py
```diff
@@ -17,6 +33,24 @@ class CopilotTextCommand(LspTextCommand, metaclass=ABCMeta):
     def want_event(self) -> bool:
         return False
 
+    def _record_telemetry(
+        self, request: str, payload: Union[CopilotPayloadNotifyAccepted, CopilotPayloadNotifyRejected]
+    ) -> None:
+        session = self.session_by_name(self.session_name)
```
timfjord, line 38: I would add
```python
if not get_setting(session, "telemetry", False):
    return
```
to avoid even getting a session if the `telemetry` set to `false`

commands.py
```diff
@@ -26,18 +60,30 @@ def run(self, edit: sublime.Edit) -> None:
...
     def run(self, _: sublime.Edit) -> None:
-        Completion(self.view).hide()
+        completion = Completion(self.view)
+        completion.hide()
+
+        completion_uuid = completion.uuid
+
+        # TODO: Currently we send the last shown completion UUID, however Copilot can
+        # suggest multiple UUID's. We need to return all UUID's which were not accepted
+        self._record_telemetry(REQ_NOTIFY_REJECTED, {"uuids": [completion_uuid]})
```
timfjord, line 85: Maybe we don't need that extra variable
```suggestion
        # TODO: Currently we send the last shown completion UUID, however Copilot can
        # suggest multiple UUID's. We need to return all UUID's which were not accepted
        self._record_telemetry(REQ_NOTIFY_REJECTED, {"uuids": [completion.uuid]})
```

ui.py
```diff
@@ -56,12 +60,15 @@ def hide(self) -> None:
 
     def show(self, region: Tuple[int, int], completions: List[CopilotPayloadCompletion], cycle: int = 0) -> None:
         cycle %= len(completions)
-        display_text = self.get_display_text(region, completions[cycle]["displayText"])
 
+        completion_uuid = completions[cycle]["uuid"]
+
+        display_text = self.get_display_text(region, completions[cycle]["displayText"])
```
timfjord, line 66: Sorry for nitpicking, but maybe we could just introduce a variable `completion` and use it here and below
```suggestion
        completion = completions[cycle]

        display_text = self.get_display_text(region, completion["displayText"])
```

commands.py
TheSecEng: We pull session settings instead of view settings. Perhaps it doesn't matter here though?

timfjord: Hm, indeed, you are right

commands.py
```diff
@@ -26,18 +58,25 @@ def run(self, edit: sublime.Edit) -> None:
...
 
         region = completion.region
-        display_text = completion.display_text or ""
-
         if not region:
             return
 
         completion.hide()
-        self.view.insert(edit, region[1], display_text)
+        self.view.insert(edit, region[1], completion.display_text)
```
timfjord, line 65: This might not work, as we reset `display_text` when we hide, so it would be an empty string here.  
We should keep `display_text = completion.display_text or ""` for the accept command

jfcherng: Wouldn't the case you say be guarded by 
```py
        if not region:
            return
```
? If not, I think we should reset region when hide.

TheSecEng: @timfjord So as per @jfcherng should we just call insert first and then hide? This would remove the empty string issue. We could also delete the key vs erase the key.

jfcherng: Why we have to reset something when doing `hide()`?

TheSecEng: Well, I think the better function name is `clear` which would clear the pop up, clear associated completion metadata. 

Either way, I think that is more of a refactor with the Completion class. I'll fix this PR and merge, and then we need to refactor Completion

timfjord: We reset the variables because we rely on theme in [the context calculation ](https://github.com/TheSecEng/LSP-copilot/blob/master/listeners.py#L34-L37)

There is `Completion.is_visible` that is responsible for that and it is checking whether both the `region` and the `dispaly_text` are present.

I think these two concerns(the UI part and the metadata) are related, and that's why they are grouped together.  
When we hide the popup we probably don't want the metadata to stay

jfcherng: > We reset the variables because we rely on them in [the context calculation ](https://github.com/TheSecEng/LSP-copilot/blob/master/listeners.py?rgh-link-date=2022-07-15T18%3A59%3A05Z#L34-L37)
> 
> There is `Completion.is_visible` that is responsible for that and it is checking whether both the `region` and the `dispaly_text` are present.

That still relies on other properties. I mean something like

```py
    def is_visible(self) -> bool:
        return bool(self._settings("is_visible") or False)
```

jfcherng: Changes are proposed on https://github.com/TheSecEng/LSP-copilot/pull/8

timfjord: We could do that, but I think we still need to keep `is_visible` in sync with `display_text`, `region` and `uuid`. And if we hide the popup we should also clean up those variables

jfcherng: I see no reason to clean them up at this moment. Except for gaining tiny memory gc.

timfjord: 🙂 Yeah, I was basically going to say, that maybe it's not a bad idea to keep them out of sync. This way, theoretically, we could even restore the last suggestion

TheSecEng:
Moved the insertion commands, and hide commands. Future refactor in #8

PR #9, opened 2022-07-16
refactor: Copilot view settings setter/getter

jfcherng:
Waiting for other PRs to be merged...

---

And maybe we want to further do 
```diff
-COPILOT_VIEW_SETTINGS_PREFIX = "copilot.completion"
+COPILOT_VIEW_SETTINGS_PREFIX = "copilot"
```
I am not sure.


utils.py
```diff
@@ -2,6 +2,21 @@
 import textwrap
 
 import sublime
+from LSP.plugin.core.typing import Any, Optional
+
+from .constants import COPILOT_VIEW_SETTINGS_PREFIX
+
+
+def get_copilot_view_setting(view: sublime.View, key: str, default: Optional[Any] = None) -> Any:
```
timfjord, line 10: I don't think that `Optiona[Any]` makes sense, as `Any` already covers `None`

jfcherng: Technically they are not the same. `Any` completely drops type checking so you can do something like `default.append(x)`. But you can't do that for `Optional[Any]` case.

jfcherng: done in [80b1d38](https://github.com/TheSecEng/LSP-copilot/pull/9/commits/80b1d38887520c68e3a70f5fe9d1a66342179ff1)

jfcherng:
I think this PR has been ready for review.

TheSecEng (approved):

PR #13, opened 2022-07-17
feat: previous/next completions

### LINKED ISSUE #6: Completions missing ability to be cycled
Copilot supports returning a list of completions, the current implementation doesn't contain logic to cycle through completions easily

jfcherng:
https://user-images.githubusercontent.com/6594915/179384384-a0e35df5-7184-4369-972b-2195edd755ee.mp4

Use `alt+[` and `alt+]` to do completion cycling.

---

Now we store following things in view settings:

- is_visible
- is_waiting
- completions (full completions payload from copilot)
- completion_index (0-based index of the current chosen completion)

---

Resolves #6

### DIFF
--- a/commands.py
+++ b/commands.py
@@ -93,7 +93,7 @@ class CopilotAskCompletionsCommand(CopilotTextCommand):
         )
 
 
-class CopilotAcceptSuggestionCommand(CopilotTextCommand):
+class CopilotAcceptCompletionCommand(CopilotTextCommand):
     @_provide_session()
     def run(self, session: Session, edit: sublime.Edit) -> None:
         completion_manager = ViewCompletionManager(self.view)
@@ -122,7 +122,7 @@ class CopilotAcceptSuggestionCommand(CopilotTextCommand):
             self._record_telemetry(session, REQ_NOTIFY_REJECTED, {"uuids": other_uuids})
 
 
-class CopilotRejectSuggestionCommand(CopilotTextCommand):
+class CopilotRejectCompletionCommand(CopilotTextCommand):
     @_provide_session()
     def run(self, session: Session, _: sublime.Edit) -> None:
         completion_manager = ViewCompletionManager(self.view)
@@ -136,12 +136,12 @@ class CopilotRejectSuggestionCommand(CopilotTextCommand):
         )
 
 
-class CopilotPreviousSuggestionCommand(CopilotTextCommand):
+class CopilotPreviousCompletionCommand(CopilotTextCommand):
     def run(self, _: sublime.Edit) -> None:
         ViewCompletionManager(self.view).show_previous_completion()
 
 
-class CopilotNextSuggestionCommand(CopilotTextCommand):
+class CopilotNextCompletionCommand(CopilotTextCommand):
     def run(self, _: sublime.Edit) -> None:
         ViewCompletionManager(self.view).show_next_completion()
 
--- a/commands.py
+++ b/commands.py
@@ -138,16 +138,12 @@ class CopilotRejectSuggestionCommand(CopilotTextCommand):
 
 class CopilotPreviousSuggestionCommand(CopilotTextCommand):
     def run(self, _: sublime.Edit) -> None:
-        completion_manager = ViewCompletionManager(self.view)
-        completion_manager.choose_previous_completion()
-        completion_manager.show()
+        ViewCompletionManager(self.view).show_previous_completion()
 
 
 class CopilotNextSuggestionCommand(CopilotTextCommand):
     def run(self, _: sublime.Edit) -> None:
-        completion_manager = ViewCompletionManager(self.view)
-        completion_manager.choose_next_completion()
-        completion_manager.show()
+        ViewCompletionManager(self.view).show_next_completion()
 
 
 class CopilotCheckStatusCommand(CopilotTextCommand):
--- a/commands.py
+++ b/commands.py
@@ -26,7 +26,7 @@ from .types import (
     CopilotPayloadSignOut,
     T_Callable,
 )
-from .ui import Completion
+from .ui import ViewCompletionManager
 from .utils import get_setting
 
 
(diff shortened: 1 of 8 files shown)


ui.py
```diff
@@ -1,75 +1,89 @@
...
-from LSP.plugin.core.typing import List, Tuple
+from LSP.plugin.core.typing import List, Optional
 
 from .types import CopilotPayloadCompletion
-from .utils import get_copilot_view_setting, reformat, set_copilot_view_setting
+from .utils import clamp, get_copilot_view_setting, reformat, set_copilot_view_setting
 
 
-class Completion:
+class ViewCompletionManager:
```
timfjord, line 12: What if we rename this to `ViewSuggestionManager`(together with `_PopupSuggestion`)?

Currently use `copilot_accept_suggestion`/`copilot_reject_suggestion` naming but under the hood we use `completion` work. I know that it doesn't really change anything, but I think it is important to stay consistent.

Alternatively, we could rename `copilot_accept_suggestion` to `copilot_accept_completion`(and `*_reject_` too) but I personally think that suggestion is more appropriate here as completion confuses with the Sublime completions

commands.py
```diff
@@ -96,32 +96,58 @@ def run(self, _: sublime.Edit) -> None:
...
-        completion = Completion(self.view)
-        completion.hide()
+        completion_manager = ViewCompletionManager(self.view)
+        completion_manager.hide()
+
+        # notify all completions as rejected
+        self._record_telemetry(
+            session,
+            REQ_NOTIFY_REJECTED,
+            {"uuids": [completion["uuid"] for completion in completion_manager.completions]},
```
timfjord, line 135: Why do we report all `uuid`s? Say, I see a suggestion and I not happy with it, why other suggestions should be also marked as reject as I might have not even seen them?

commands.py
```diff
@@ -96,32 +96,58 @@ def run(self, _: sublime.Edit) -> None:
...
+        # notify the current completion as accepted
+        self._record_telemetry(session, REQ_NOTIFY_ACCEPTED, {"uuid": completion["uuid"]})
 
-        # TODO: When a suggestion is accept, we need to send a REQ_NOTIFY_REJECTED
-        # request with all other completions which weren't accepted
-        self._record_telemetry(session, REQ_NOTIFY_ACCEPTED, {"uuid": completion.uuid})
+        # notify all other completions as rejected
+        other_uuids = [completion["uuid"] for completion in completion_manager.completions]
+        other_uuids.remove(completion["uuid"])
+        if other_uuids:
```
timfjord, line 121: Do we really need to mark other `uuid`s as rejected?

commands.py
```diff
@@ -96,32 +96,58 @@ def run(self, _: sublime.Edit) -> None:
...
+
 
-        # TODO: Currently we send the last shown completion UUID, however Copilot can
-        # suggest multiple UUID's. We need to return all UUID's which were not accepted
-        self._record_telemetry(session, REQ_NOTIFY_REJECTED, {"uuids": [completion.uuid]})
+class CopilotPreviousSuggestionCommand(CopilotTextCommand):
+    def run(self, _: sublime.Edit) -> None:
+        completion_manager = ViewCompletionManager(self.view)
+        completion_manager.choose_previous_completion()
+        completion_manager.show()
```
timfjord, line 143: Maybe we should call `show` directly in `choose_previous_completion`/`choose_next_completion`?

We could even rename `choose_next_completion` to just `next_completion` since `show` will be called under the hood.

So the command will be as simple as `ViewCompletionManager(self.view).previous_completion()`

plugin.py
```diff
@@ -175,4 +175,8 @@ def _on_get_completions(
         if not completions:
             return
 
-        Completion(view).show(region, completions)
+        preprocess_completions(view, completions)
+
+        set_copilot_view_setting(view, "completions", completions)
+        set_copilot_view_setting(view, "completion_index", 0)
```
timfjord, line 181: I think we should handle this in the `ViewCompletionManager` as it looks more like a completion concern.

For example, we could have `completions` and `index` params in the `show` method but make them `None` by default. When they are passed then we update the view settings before getting them.

Because as of now it is a bit confusing, in order to make the `ViewCompletionManager` work you need to set something outside of it so we wouldn't be able to test this class properly without know all this prerequirements

ui.py
```diff
@@ -1,75 +1,89 @@
...
-        PopupCompletion(self.view, region, display_text).show()
+        :param      value:     The wanted new completion index.
+        :param      do_clamp:  Should the `value` be clamped if it's
+                               out-of-bound? Otherwise, treat it as cyclic.
+        """
+        completions_cnt = len(self.completions)
+        if completions_cnt:
+            value = clamp(value, 0, completions_cnt - 1) if do_clamp else value % completions_cnt
+        else:
+            value = 0
```
timfjord, line 82: I think the `completion_index` should be always within the range, so `do_clamp` should be always on.

To simplify the implementation we could still use `%`.
But when the `value` is negative we simply use `completions_cnt + value` as the final index before the `%` thing.  
This way the index will be always within the range

commands.py
jfcherng: @TheSecEng since this is his comment TODO.

commands.py
jfcherng: @TheSecEng since this is his comment TODO.

ui.py
jfcherng: There is a reason why ST introduce `auto_complete_cycle` setting.

commands.py
TheSecEng: I'll analyze VSCode copilot again, but from my brief understand, they record which suggestions where selected as a ways to determine more context on your codebase. 

So when you accept one, it implies the others weren't accepted. 

I'll double check the code again but I believe this is their intent

commands.py
TheSecEng: See above comment

commands.py
jfcherng: LGTM. Will do later.

ui.py
jfcherng: I actually thought to do it in a reversed way to sync with the server protocol naming. https://github.com/TheSecEng/LSP-copilot/blob/8520ae9033d2c4a4ad117c227630cf172c763d02/constants.py#L11-L13

commands.py
jfcherng: Move discussion to https://github.com/TheSecEng/LSP-copilot/pull/13#discussion_r922831886

commands.py
jfcherng: Done in [262207e](https://github.com/TheSecEng/LSP-copilot/pull/13/commits/262207e62bc78731d422214cc33f7114a3c3829c) though I use different method naming.

ui.py
timfjord: I think both work, unless we stay consistent

plugin.py
jfcherng: Like [d5679c3](https://github.com/TheSecEng/LSP-copilot/pull/13/commits/d5679c30646608e10f5c93d44b38b76a77b0d400) ?

ui.py
timfjord: Hm, I wasn't aware of `auto_complete_cycle`. To be honest, I've been looking for this setting and even thought there was a bug in Sublime 🙂
Yes, it makes sense to respect this setting then 👍🏻

ui.py
```diff
@@ -1,75 +1,104 @@
...
-        index = raw_display_text.find(following_text)
+    def current_completion(self) -> Optional[CopilotPayloadCompletion]:
+        """The current chosen `completion`."""
+        return self.completions[self.completion_index] if self.completions else None
+
+    def show_previous_completion(self) -> None:
+        """Set `completion_index` to be for the previous completion."""
+        self.show(
+            None,
+            self.completion_index - 1,
```
timfjord, line 40: Let's use keyword arguments to avoid `None`, e.g. `completion_index=...`

jfcherng: Done in [c39d438](https://github.com/TheSecEng/LSP-copilot/pull/13/commits/c39d4389eb499ccfbff25b2792ff9608569f7ed4)

ui.py
```diff
@@ -1,75 +1,102 @@
...
+        do_clamp: bool = True,
+    ) -> None:
+        """Show Copilot's completion popup."""
+        # update completions
+        if completions is not None:
+            set_copilot_view_setting(self.view, "completions", completions)
+        # update completion index
+        if completion_index is not None:
+            set_copilot_view_setting(self.view, "completion_index", completion_index)
+        self._tidy_completion_index(do_clamp=do_clamp)
```
timfjord, line 69: Maybe just `(do_clamp)`?

timfjord:
As for the UI, maybe we should remove the right border in the prev button and the left border in the next button. This way they will be like one button?

ui.py
jfcherng: Done in [0fe023f](https://github.com/TheSecEng/LSP-copilot/pull/13/commits/0fe023fa1091bb206751cc838017dbe61d3373bd)

jfcherng:
> As for the UI, maybe we should remove the right border in the prev button and the left border in the next button. This way they will be like one button?

image
[3ff03dc](https://github.com/TheSecEng/LSP-copilot/pull/13/commits/3ff03dcd401ace38e7652ddcf61e8f0912e8c1db)

timfjord:
Not bad 👍🏻

ui.py
jfcherng: [refactor: rename "suggestion" to "completion"](https://github.com/TheSecEng/LSP-copilot/pull/13/commits/71be8312ea75ff5f24bedff44fc21a0e4aa691f3) 

@timfjord This change incudes command names so your custom keybindings have to be renamed as well.

TheSecEng (approved):

commands.py
jfcherng: Let's merge this PR. If telemetry parts need to be changed, we can do it in another PR later.

PR #18, opened 2022-07-18
feat: panel completions

TheSecEng:
WIP for Panel Completions

@jfcherng @timfjord Feel free to murder this PR. I hacked it together but haven't done much refactoring

image


TheSecEng:
Great contributions @jfcherng one thing that does work right now is the actual completion!

Each link needs a index to reference which panel completion was accepted and then we handle the insertion just like any other completion.

TheSecEng:
@jfcherng I am done for the night. 

A few things

1. Panel completions should close any current Popup Completions
2. Panel Completions should be cleared and closed if a solution is accepted 
3. Panel Completions should be clear if the Panel Completions sheet is closed
4. Completion Insertion isn't working (ST issue maybe?)

jfcherng:
I noticed that the inserted completion is somehow incomplete... with a comparison to the one in the panel.

commands.py
```diff
@@ -93,11 +107,35 @@ def run(self, _: sublime.Edit) -> None:
...
+    def run(self, view_id: int, completion_index: int) -> None:
+        target_view = first(all_st_views(), lambda view: view.id() == view_id)
+        if not target_view:
+            return
+        target_view.run_command("copilot_accept_panel_completion", {"completion_index": completion_index})
+
+
+class CopilotAcceptPanelCompletionCommand(CopilotTextCommand):
+    def run(self, edit: sublime.Edit, completion_index: int) -> None:
+        completion_manager = ViewCompletionManager(self.view)
```
timfjord, line 125: I think we should not use the `ViewCompletionManager` class for panel completion, instead, we should add `ViewPanelCompletionManager` and encapsulate all panel completion-related logic there

commands.py
```diff
@@ -141,6 +179,33 @@ def run(self, session: Session, _: sublime.Edit) -> None:
...
+    @_provide_session()
+    def run(self, session: Session, _: sublime.Edit) -> None:
+        params = prepare_completion_request(view=self.view)
+        if not params:
+            return
+
+        panel_id = "copilot://{}".format(self.view.id())
+        params["panelId"] = panel_id
+
+        set_copilot_view_setting(self.view, "panel_id", panel_id)
```
timfjord, line 192: If we introduce `ViewPanelCompletionManager`, all this could be part of it

commands.py
```diff
@@ -93,11 +107,35 @@ def run(self, _: sublime.Edit) -> None:
         debounced(
             functools.partial(plugin.request_get_completions, self.view),
             FEATURES_TIMEOUT,
-            lambda: not get_copilot_view_setting(self.view, "is_waiting", False),
+            lambda: not get_copilot_view_setting(self.view, "is_waiting_completions", False),
```
timfjord, line 110: I think we should hide this logic in the plugin (e.g. `plugin.is_waiting_completions()`)

plugin.py
```diff
@@ -128,40 +136,49 @@ def is_valid_for_view(self, view: sublime.View) -> bool:
     def _handle_log_message_notification(self, payload: CopilotPayloadLogMessage) -> None:
         pass
 
+    @notification_handler(NTFY_PANEL_SOLUTION)
+    def _handle_panel_solution_notification(self, payload: CopilotPayloadPanelSolution) -> None:
```
timfjord, line 140: Why do we need this? I mean cannot we handle all this in the `NTFY_PANEL_SOLUTION_DONE`?

plugin.py
```diff
@@ -128,40 +136,49 @@ def is_valid_for_view(self, view: sublime.View) -> bool:
...
         ViewCompletionManager(view).hide()
 
         session = self.weaksession()
-        syntax = view.syntax()
         sel = view.sel()
-        if not (self.get_has_signed_in() and session and syntax and len(sel) == 1):
+        if not (self.get_has_signed_in() and session and len(sel) == 1):
             return
 
-        cursor = sel[0]
```
timfjord, line 144: Just a note regarding the idea I shared above.

Instead of handling everything with utility functions(like `prepare_completion_request`) we could go OOP and have something like `Request(view).perform()`

We could introduce a base abstract class interface and require `perform` and `handle` methods to be implemented for every `BaseRequest` subclass. 

It is not critical for this PR I think, and we should probably refactor this after we merge it

plugin.py
jfcherng: It would take quite a long time. Solutions are sent one by one once Copilot figures out one of them.

plugin.py
jfcherng: > It is not critical for this PR I think, and we should probably refactor this after we merge it

👍

ui.py
```diff
@@ -230,3 +252,165 @@ def show(self) -> None:
...
+        syntax = self.view.syntax() or sublime.find_syntax_by_name("Plain Text")[0]
+        view_id = int(remove_prefix(get_copilot_view_setting(self.view, "panel_id", -1), "copilot://"))
+        completions = self._synthesize(self.completion_manager.panel_completions)
+        return self.COMPLETION_TEMPLATE.format(
+            index=len(self.completion_manager.panel_completions),
+            total_solutions=get_copilot_view_setting(self.view, "panel_completion_target_count", 0),
+            sections="\n\n<hr>\n".join(
+                self.COMPLETION_SECTION_TEMPLATE.format(
+                    header_items=" &nbsp;".join(self.completion_header_items(completion, view_id, index)),
+                    score=completion["score"],
```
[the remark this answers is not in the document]
jfcherng: > What if we render the score in the same line with the Accept/Reject button

I think it's done in that way already.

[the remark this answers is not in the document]
timfjord: Cool, it is looking great 👍🏻

ui.py
```diff
@@ -230,3 +252,164 @@ def show(self) -> None:
...
+                sublime.command_url(
+                    "copilot_accept_panel_completion_shim",
+                    {"view_id": view_id, "completion_index": index},
+                )
+            ),
+            "(Mean Probability: {})".format(completion["score"]),
+        ]
+
+    def open(self) -> None:
+        # TODO: show this side-by-side?
```
timfjord, line 348: Probably, eventually, we could make this configurable

VSCode opens in a split window to the right

TheSecEng: Yeah, I think that should be in this PR tbh. Doesn't make sense to hide the view your working in. 

Btw I made you a contributor to this repo.

utils.py
```diff
@@ -49,14 +65,84 @@ def get_setting(session: Session, key: str, default: Optional[Union[str, bool, L
...
+    }
+
+
 def preprocess_completions(view: sublime.View, completions: List[CopilotPayloadCompletion]) -> None:
     for completion in completions:
         completion["positionSt"] = view.text_point(
             completion["position"]["line"],
             completion["position"]["character"],
         )
+        completion["rangeSt"] = (
```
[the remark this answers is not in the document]
jfcherng: Done in fafb4919d829bd83a5475618064e904120a8e484

commands.py
jfcherng: Done in [d641cff](https://github.com/TheSecEng/LSP-copilot/pull/18/commits/d641cff22ca9270ea70a9b2de1d34c2ed077b7b4)

types.py
```diff
@@ -1,7 +1,10 @@
-from LSP.plugin.core.typing import Any, Callable, List, Literal, TypedDict, TypeVar
+from LSP.plugin.core.typing import Any, Callable, List, Literal, Tuple, TypedDict, TypeVar
 
 T_Callable = TypeVar("T_Callable", bound=Callable[..., Any])
 
+StPoint = int
```
timfjord, line 5: This `StPoint` looks like `Saint Point` 😄 
Maybe we should rather call it `SublimeTextPoint`

types.py
```diff
@@ -1,7 +1,10 @@
...
+StRegionTuple = Tuple[int, int]
```
timfjord, line 6: I think technically it should be
```suggestion
StRegionTuple = Tuple[StPoint, StPoint]
```

ui.py
```diff
@@ -230,3 +231,230 @@ def show(self) -> None:
...
+    def panel_id(self) -> str:
+        return "copilot://{}".format(self.view.id())
+
+    @property
+    def completions(self) -> List[CopilotPayloadPanelSolution]:
+        """All `completions` in the view. Note that this is a copy."""
+        return get_copilot_view_setting(self.view, "panel_completions", [])
+
+    @completions.setter
+    def completions(self, value: List[CopilotPayloadPanelSolution]) -> None:
```
timfjord, line 301: 👍🏻

jfcherng:
Gonna sleep.

plugin.py
```diff
@@ -62,15 +59,19 @@ def __init__(self, session: "weakref.ref[Session]") -> None:
...
+                cm.is_visible = False
+                cm.is_waiting = False
+
+                pcm = ViewPanelCompletionManager(view)
+                pcm.is_waiting = False
 
     def on_ready(self, api: ApiWrapperInterface) -> None:
         def on_check_status(result: CopilotPayloadSignInConfirm, failed: bool) -> None:
-            self.set_has_signed_in(result.get("status") == "OK")
+            self.set_has_signed_in(result["status"] == "OK")
```
TheSecEng, line 74: There is no guarantee that the message will have a status here and we don't handle it properly with this approach. One example is if there is a `self-signed` certificate in the chain it will return the following:

```json
{'message': 'Request checkStatus failed with message: self signed certificate in certificate chain', 'code': -32603}
```

types.py
TheSecEng: Saint Point, a lovely fellow

TheSecEng:
@jfcherng should we open a new window and then do side by side for completion panel? Once closed or accepted or selection changed, we can close and restore to the previous state. 

Not sure the best path forward.

plugin.py
jfcherng: > There is no guarantee that the message will have a status here and we don't handle it properly with this approach. One example is if there is a `self-signed` certificate in the chain it will return the following:
> 
> ```json
> {
> 	"message": "Request checkStatus failed with message: self signed certificate in certificate chain", 	 
>         "code": -32603
> }
> ```

But I think then they can't use this plugin?

jfcherng:
> should we open a new window and then do side by side for completion panel?

Sounds good for me. Open a dedicated window with the same buffer (not sure whether this is possible. same underlying buffer but different view) and side-by-side panel completion. After it's done, we probably close the whole window.

timfjord:
> 1. Create a new window

Do we really need a new window?  
It looks like we need a new view to open it side by side

jfcherng:
> > 1. Create a new window
> 
> Do we really need a new window? It looks like we need a new view to open it side by side

Depends on whether the user wants to wait for panel completion.

jfcherng:
Another tricky question: What should happen if the user changes his cursor position before he accepts a panel completion?

plugin.py
TheSecEng: Won't work

TheSecEng:
Aligning with VSCode

image


image

timfjord:
It looks like minihtml supports `right` and `position: static;`
So theoretically, it could be
```css
position: static;
left: auto;
right: 0.5rem; // or just 0 
```

plugin.py
```diff
@@ -128,40 +128,49 @@ def is_valid_for_view(self, view: sublime.View) -> bool:
...
+        view = ViewPanelCompletionManager.find_view_by_panel_id(payload["panelId"])
+        if not view:
+            return
+
+        preprocess_panel_completions(view, [payload])
+
+        completion_manager = ViewPanelCompletionManager(view)
+        completions = completion_manager.completions
+        completions.append(payload)
+        completion_manager.completions = completions
```
timfjord, line 142: Maybe we could introduce `completion_manager.append_completion(payload)`?

ui/panel_completion.py
```diff
@@ -0,0 +1,294 @@
...
+
+    # -------------- #
+    # normal methods #
+    # -------------- #
+
+    def __init__(self, view: sublime.View) -> None:
+        self.view = view
+
+    def get_completion(self, index: int) -> Optional[CopilotPayloadPanelSolution]:
+        return next(iter(self.completions[index : index + 1]), None)
```
timfjord, line 82: Wouldn't 
```suggestion
        try:
            return self.completions[index]
        except IndexError:
            return None    
```
be simpler?

jfcherng: i don't care. change that if you preferred.

timfjord: Just thought it would be simpler this way. But it is not critical at all

plugin.py
jfcherng: sounds good.

jfcherng: Done in 0550292

jfcherng:
I noticed that at this moment, the only "correct" way to close a completion panel is accepting a completion. Otherwise, the original layout won't be restored.

TheSecEng:
> I noticed that at this moment, the only "correct" way to close a completion panel is accepting a completion. Otherwise, the original layout won't be restored. Unfortunately, there is no listener or callback for a closed sheet.

Yes, this is something that is a pain. What about a recommendation from Facelessuser. We create a view and then create a massive phantom in the view?

TheSecEng:
Kinda of on, I feel like the `on_close` lamdba for sheets should also trigger for when closed like a view. Do sheets get assigned a view as well?

commands.py
```diff
@@ -51,9 +64,23 @@ def wrap(self, *arg, **kwargs) -> Any:
     return decorator
 
 
-class CopilotTextCommand(LspTextCommand, metaclass=ABCMeta):
+class CopilotCommandBase(metaclass=ABCMeta):
     session_name = PACKAGE_NAME
+    requirement = REQUIRE_SIGN_IN
```
timfjord, line 69: I've been thinking about this API here.

Bitflags are useful when we need to support multiple conditions, but in our case, most likely we would never use anything else except `SIGN_IN`/`NOT_SIGN_IN`.  
So maybe we should rather go with a simple class variable:
```python
requires_sign_in: Optional[bool] = None
```
and when it `self.requires_sign_in is None` we don't do anything(even calling `CopilotPlugin.get_has_signed_in`), otherwise, we could do
```python
return True if self.requires_sign_in is None else self.requires_sign_in == CopilotPlugin.get_has_signed_in()
```
It is not critical, just thought it would make this API more obiouvs

plugin.py
```diff
@@ -62,15 +65,18 @@ def __init__(self, session: "weakref.ref[Session]") -> None:
...
-        # ST persists view setting after getting closed so we have to reset some status
-        for window in sublime.windows():
-            for view in window.views(include_transient=True):
-                erase_copilot_view_setting(view, "is_visible")
-                erase_copilot_view_setting(view, "is_waiting")
+        # ST persists view settings after getting closed so we have to reset some status
+        for view in all_views():
+            cm = ViewCompletionManager(view)
+            cm.is_visible = False
+            cm.is_waiting = False
```
timfjord, line 72: Maybe we should introduce the `reset` method: `ViewCompletionManager(view).reset()` (for the `ViewPanelCompletionManager` too)

plugin.py
```diff
@@ -128,40 +133,47 @@ def is_valid_for_view(self, view: sublime.View) -> bool:
...
+        completion_manager.append_completion(payload)
+        completion_manager.update()
+
+    @notification_handler(NTFY_PANEL_SOLUTION_DONE)
+    def _handle_panel_solution_done_notification(self, payload) -> None:
+        view = ViewPanelCompletionManager.find_view_by_panel_id(payload["panelId"])
+        if not view:
+            return
+
+        ViewPanelCompletionManager(view).is_waiting = False
```
timfjord, line 154: With `reset`, it could be more readable `ViewPanelCompletionManager(view).reset()`

ui/completion.py
```diff
@@ -6,47 +6,77 @@
...
         return get_copilot_view_setting(self.view, "completion_index", 0)
 
+    @completion_index.setter
+    def completion_index(self, value: int) -> None:
+        """The index of the current chosen completion."""
+        do_clamp = not self.view.settings().get("auto_complete_cycle", False)
+        set_copilot_view_setting(
+            self.view,
+            "completion_index",
+            self._tidy_completion_index(value, do_clamp=do_clamp),
```
timfjord, line 58: It is outside of the scope of this PR, but since we are refactoring some stuff, maybe we don't need this `do_clamp` param. Instead, `do_clamp = not self.view.settings().get("auto_complete_cycle", False)` could be hardcoded in the `_tidy_completion_index`.  
To be honest, I would get rid of it at all, as it is the responsibility of this setter to handle the index

ui/completion.py
```diff
@@ -180,7 +210,7 @@ def __init__(self, view: sublime.View) -> None:
 
     @property
     def popup_content(self) -> str:
-        syntax = self.view.syntax() or sublime.find_syntax_by_name("Plain Text")[0]
+        syntax = self.view.syntax() or PLAIN_TEXT_SYNTAX
```
timfjord, line 213: Maybe `view.syntax() or PLAIN_TEXT_SYNTAX` can be extracted to a helper method, `get_view_syntax`.  
We just use it quite often

ui/panel_completion.py
```diff
@@ -0,0 +1,279 @@
...
+            window=window,
+            name="Panel Completions",
+            contents=self.completion_content,
+            md=True,
+            css=self.CSS,
+            wrapper_class=self.CSS_CLASS_NAME,
+        )
+        self.completion_manager.sheet_id = sheet.id()
+
+    def _open_in_side_by_side(self, window: sublime.Window) -> None:
```
timfjord, line 270: When I suggested `side_by_side` I mean the ST4 side-by-side thing(when two views are opened together), this way we don't need t worry about restoring the layout.

utils.py
```diff
@@ -49,14 +78,101 @@ def get_setting(session: Session, key: str, default: Optional[Union[str, bool, L
     return value
 
 
+def message_dialog(msg: str, *, console: bool = False) -> None:
```
timfjord, line 81: What if we delegate formatting of the msg to this helper:
```python
message_dialog(msg: str, *, console: bool = False, **kwargs) -> None:
    msg = msg.format(**kwargs)
    # ...
```
this way, we could use it like `message_dialog('Co{what}', what="pilot")`

timfjord (commented):
The RR looks great 💪🏻
Added a couple of not critical suggestions

commands.py
jfcherng: How can `requires_sign_in: Optional[bool]` represents `REQUIRE_NOT_SIGN_IN`?

jfcherng: That makes the behavior more vague imho... How do you represents `REQUIRE_NOT_SIGN_IN` without introducing another variable?

ui/panel_completion.py
jfcherng: What you mean is a transient view. I don't think we can create one via existing API?

jfcherng:
From some @TheSecEng's discussion, I think we will drop the HtmlSheet solution given that there is no way to fire a callback when the user close a sheet. View + (big) phantom will be a replacement.

plugin.py
jfcherng: Done in [a168798](https://github.com/TheSecEng/LSP-copilot/pull/18/commits/a16879844cf092d263701cb321475b3e143e8044)

plugin.py
jfcherng: Done in [a168798](https://github.com/TheSecEng/LSP-copilot/pull/18/commits/a16879844cf092d263701cb321475b3e143e8044)

ui/completion.py
jfcherng: Done in [50cb36d](https://github.com/TheSecEng/LSP-copilot/pull/18/commits/50cb36dd15f1300424a21df184af5bda2e6e1b9f)

commands.py
timfjord: My idea was to use `False` to ensure sign out, `True` to ensure sign in, `None` to bypass the check altogether

ui/panel_completion.py
timfjord: Here is how it is handled in LSP https://github.com/sublimelsp/LSP/blob/5fc0db5640d1b6515ad4a6f58f6797613e012e06/plugin/locationpicker.py#L24  
Basically `sublime.ADD_TO_SELECTION | sublime.SEMI_TRANSIENT`

jfcherng: That should be an idea solution I think. @TheSecEng

commands.py
jfcherng: I feel it's a little vague that `requires_sign_in = False` means `require not sign in` or `not require sign in`.

ui/completion.py
jfcherng: Done in [cf74887](https://github.com/TheSecEng/LSP-copilot/pull/18/commits/cf748870347d6f835ce10102d5ad830db9b238ad)

commands.py
timfjord: Agree 👍🏻

plugin.py
```diff
@@ -128,40 +131,47 @@ def is_valid_for_view(self, view: sublime.View) -> bool:
...
```
timfjord, line 152:
```suggestion
        ViewPanelCompletionManager(view).reset()
```

utils.py
jfcherng: image
Copilot agrees.

jfcherng: Done in [eb68648](https://github.com/TheSecEng/LSP-copilot/pull/18/commits/eb686481a56a6a18adcd73d375c28e6e6c35d504)

ui/panel_completion.py
TheSecEng: This works for existing files and `window.open_file` this doesn't work for `window.new_file`

TheSecEng: Also the phantom approach is a horrible UX. It flickers every update. Either of you want to take a stab at it?

TheSecEng: > This works for existing files and `window.open_file` this doesn't work for `window.new_file`

Or altleazst I can't get it to wrk

jfcherng: ok. let's create a dummy file for opening 🤣

TheSecEng:
https://github.com/TheSecEng/LSP-copilot/tree/feat/gpcTransient

This is using a `HtmlSheet` with the TRANSIENT flag, removing the Tab which helps remove the tab close issues. @jfcherng If you want to see about hooking into the close command, this should be a suitable solution

jfcherng:
Hmm... we seem to have to monitor both the view and the corresponding sheet...

- If the view gets closed first, the sheet has to be close too and restore layout.
- If the sheet gets closed first, restore layout.

TheSecEng:
> Hmm... we seem to have to monitor both the view and the corresponding sheet...
> 
> * If the view gets closed first, the sheet has to be closed too and restore layout.
> * If the sheet gets closed first (by `close` / `close_file` command), restore layout.
> 
> OK. Both done on your branch. But I would like to see/tweak the view + big phantom method. :)

https://github.com/TheSecEng/LSP-copilot/tree/feat/gpcPhantoms

Closing isn't working yet. Not sure how to approach it properly

TheSecEng:
Closing is working properly on the Phantoms branch

TheSecEng:
https://user-images.githubusercontent.com/32599364/180479463-a0842b20-dc04-4a9c-84d4-a5e54bf5d40d.mp4

TheSecEng:
Hmm right now the only thing that isn't working is the the same phantom might appear twice. Need logic to ensure only unique phantoms are added

jfcherng:
Hmm... also the minimap makes it looks bad.

jfcherng:
> Hmm right now the only thing that isn't working is the the same phantom might appear twice. Need logic to ensure only unique phantoms are added

Fixed by closing the view and create a new one. But now I vote +1 for the HtmlSheet.

TheSecEng:
> > Hmm right now the only thing that isn't working is the the same phantom might appear twice. Need logic to ensure only unique phantoms are added
> 
> 
> 
> Fixed by closing the view and create a new one. But now I vote +1 for the HtmlSheet.

Yeah, based on so many factors, I think the HtmlSheet is the best approach, and hopefully we can get ST to implement a evenntlistener for the sheets

jfcherng:
If @timfjord doesn't have other suggestion or something want to try, let's merge `feat/gpcTransient` to this PR.

timfjord:
What if we make the header a bit smaller, I think it uses too much space
image
Ideally, the close button should be from the right, but I couldn't find a way how to move it there (can play with it later once we merge this)

timfjord:
In general, I think it is looking really great. Good job folks 👍🏻. 
We could think about how to address the limitations in a follow-up PRs.

For me, the biggest issue is that you cannot select and copy the text from the panel, so maybe, we could literally render the text as it is and add the headers as phantoms. But again, we can play with it once we have this merged as it is more than enough for the first version.

jfcherng:
image

I think people care about the whole process has been ended or not. They don't care how many duplicates.

commands.py
```diff
@@ -134,7 +134,7 @@ def run(self, edit: sublime.Edit, completion_index: int) -> None:
         # We don't have to care whether it's an inline completion or not.
         source_line_region = self.view.line(sublime.Region(*completion["region"]))
         self.view.erase(edit, source_line_region)
-        self.view.insert(edit, source_line_region.begin(), completion["completionText"])
+        self.view.insert(edit, source_line_region.begin(), completion["displayText"])
```
jfcherng, line 161: The tricky part is that I think `displayText` shows how it should looks like for the "whole" block. Not only those lines after the current line. While `completionText` is those lines after the current line, which is for insertion imo.

timfjord:
I think it is a good idea, we could even add show a loading icon(minihtml supports that)

commands.py
TheSecEng: So you have an example screenshot of this ? I've never experienced it

jfcherng: Hmm... not sure what is supposed to do in this case honestly.
image

TheSecEng: That is a weird case, we should prevent panelcompletions if the region / line where the cursor is empty as a work around until I see how VSCode handles these

jfcherng:
> gif(I think minihtml supports that)

yes, but not animated

jfcherng:
I've merged `feat/gpcTransient` into this PR since `feat/gpcPhantoms` is not likely to be a solution at this moment.

timfjord:
Hm, then I guess we could say `Synthesizing ...` and add `(Done)` like you suggested

jfcherng:
> Hm, then I guess we could say `Synthesizing ...` and add `(Done)` like you suggested

image
image

TheSecEng:
> > Hm, then I guess we could say `Synthesizing ...` and add `(Done)` like you suggested
> 
> like these? image image

This looks good. I liked the hour glass icon as well. But I will leave that up to you

timfjord:
Maybe add `...` in the end we of the sentence?

jfcherng:
image

You may see a black/white emoji. It's colorful because I have a colorful emoji system font.

TheSecEng (commented):
Looks good

timfjord:
It looks nice 👍🏻

jfcherng:
If there is no other issues, let's (squash and) merge this PR.

PR #26, opened 2022-07-26
feat: hook to auto_complete command

timfjord:
When I started working on this PR my idea was to only add `hook_to_auto_complete_command` settings so the copilot completion appears after clicking `ctrl+space`.  
But then I realised that our current approach of handling the `modified` event breaks the undo functionality(it undos symbol by symbols instead of changes) so this PR also fixes that

### DIFF
--- a/LSP-copilot.sublime-settings
+++ b/LSP-copilot.sublime-settings
@@ -12,6 +12,7 @@
 	"initializationOptions": {},
 	"settings": {
 		"auto_ask_completions": true,
+		"hook_to_auto_complete_command": false,
 		"debug": false,
 		"telemetry": false
 	},
--- a/listeners.py
+++ b/listeners.py
@@ -55,7 +55,7 @@ class ViewEventListener(sublime_plugin.ViewEventListener):
             return test(CopilotPlugin.get_account_status().is_authorized)
         return None
 
-    def on_post_text_command(self, command_name: str, args: dict):
+    def on_post_text_command(self, command_name: str, args: Optional[Dict[str, Any]]) -> None:
         if command_name != "auto_complete":
             return
 
--- a/listeners.py
+++ b/listeners.py
@@ -1,5 +1,9 @@
+import functools
+
 import sublime
 import sublime_plugin
+from LSP.plugin import Session
+from LSP.plugin.core.types import FEATURES_TIMEOUT, debounced
 from LSP.plugin.core.typing import Any, Dict, Optional, Tuple
 
 from .plugin import CopilotPlugin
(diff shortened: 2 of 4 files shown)


sublime-package.json
```diff
@@ -24,6 +24,11 @@
                       "description": "Auto ask the server for completions. Otherwise, you have to trigger it manually.",
                       "type": "boolean"
                     },
+                    "hook_to_auto_complete_command": {
+                      "default": true,
+                      "description": "Ask the server for completions when the auto_complete command is called.",
```
jfcherng, line 29:
```suggestion
                      "markdownDescription": "Ask the server for completions when the `auto_complete` command is called.",
```

listeners.py
```diff
@@ -41,6 +55,15 @@ def test(value: Any) -> Optional[bool]:
             return test(CopilotPlugin.get_account_status().is_authorized)
         return None
 
+    def on_post_text_command(self, command_name: str, args: dict):
```
jfcherng, line 58:
```suggestion
    def on_post_text_command(self, command_name: str, args: Optional[Dict[str, Any]]) -> None:
```

LSP-copilot.sublime-settings
```diff
@@ -12,6 +12,7 @@
 	"initializationOptions": {},
 	"settings": {
 		"auto_ask_completions": true,
+		"hook_to_auto_complete_command": false,
```
jfcherng, line 15: I guess this is defaulted to `true`, as specified in `sublime-package.json`.

listeners.py
```diff
@@ -41,6 +55,15 @@ def test(value: Any) -> Optional[bool]:
...
         return None
 
+    def on_post_text_command(self, command_name: str, args: Optional[Dict[str, Any]]) -> None:
+        if command_name != "auto_complete":
+            return
+
+        _, session = self._get_session()
+
+        if session and get_setting(session, "hook_to_auto_complete_command"):
+            self.view.run_command("copilot_ask_completions")
```
timfjord, line 65: I just realsied, that I can do `plugin.request_get_completions(self.view)` directly here

LSP-copilot.sublime-settings
timfjord: Hm, there is a typo in `sublime-package.json` as my idea was to make it false by default


But I am happy to make it `true`

listeners.py
jfcherng: Seems ok. `request_get_completions` has authorization check.

LSP-copilot.sublime-settings
jfcherng: Since it's configurable, I think the default is not really important. Just to sync two side either way.

timfjord: Made it `false`

listeners.py
```diff
@@ -8,17 +12,27 @@
...
         if not plugin:
-            return
+            return None, None
 
         session = plugin.weaksession()
         if not session:
-            return
+            return plugin, None
+
+        return plugin, session
```
jfcherng, line 24: Maybe just

```py
return plugin, plugin.weaksession()
```

timfjord: even shorter, I like it 👍🏻

jfcherng (approved):

TheSecEng (approved):
LTGM

PR #27, opened 2022-07-26
feat: add phantom completion

timfjord:
image

- [x] hide phantoms `on_selection_modified_async`
- [x] handle multi-line display text
- [x] make the completion method configurable
- [x] clean up the phantoms dict when a view is closed
- [x] try `line-height: 0` for the empty block phantom we use to keep the cursor from the left

TODOs:
- [ ] update the currently displayed phantom(instead of making a new request) when typing characters that match the beginning of the currently displayed phantom (this should work even when `auto_ask_completions` is set to `false`)
- [ ] split the display text into regions and display with multiple phantoms
- [ ] fix corner cases when the cursor jumps when the completion at the end of the line
- [ ] do not ask for completion when saving a view with `lsp_format_on_save`
- [ ] do not ask for completion after `undo`/`redo`
- [ ] [optional] override the background colour for the block phantom and make it the same as the background colour of the editor (try [this workaround](https://stackoverflow.com/a/52440845/1078179) to make it full width), this way it will be different from the `line_highlight` colour


jfcherng:
@TheSecEng has something similar or superior iirc

TheSecEng:
@timfjord this is what I was working on in a local branch.

image


Is this something that is worth pursuing?

plugin/ui/completion.py
```diff
@@ -83,6 +84,7 @@ def hide(self) -> None:
         # prevent from hiding other plugin's popup
         if self.is_visible:
             _PopupCompletion.hide(self.view)
+            _PhantonCompletion.hide(self.view)
```
jfcherng, line 87: typo: Phanto`m`

timfjord: 🤦🏻

timfjord:
> @timfjord this is what I was working on in a local branch.
> 
> image
> 
> Is this something that is worth pursuing?

Absolutely @TheSecEng 

I just started on the implementation and there are a lot of things to do to make the integration really nice.  
So if you have something please share it, so we could then combine some ideas

plugin/ui/completion.py
```diff
@@ -118,7 +121,22 @@ def _tidy_completion_index(self, index: int) -> int:
...
+    def __init__(self, view: sublime.View) -> None:
+        self.view = view
+        self.completion_manager = ViewCompletionManager(view)
+
+    @abstractmethod
+    def show(self) -> None:
+        pass
+
+    @classmethod
+    @abstractmethod
```
jfcherng, line 134:
```suggestion
    @abstractclassmethod
```

plugin/ui/completion.py
```diff
@@ -118,7 +121,22 @@ def _tidy_completion_index(self, index: int) -> int:
         return clamp(index, 0, completions_cnt - 1)
 
 
-class _PopupCompletion:
+class BaseCompletion(metaclass=ABCMeta):
```
jfcherng, line 124:
```suggestion
class BaseCompletion(ABC):
```

And `from abc import ABC`

jfcherng: Sorry, forget we are at py33.

plugin/ui/completion.py
jfcherng, line 87:
```suggestion
            _PhantomCompletion.hide(self.view)
```

plugin/ui/completion.py
```diff
@@ -105,6 +107,7 @@ def show(
             return
 
         _PopupCompletion(self.view).show()
+        _PhantonCompletion(self.view).show()
```
jfcherng, line 110:
```suggestion
        _PhantomCompletion(self.view).show()
```

plugin/ui/completion.py
```diff
@@ -259,3 +273,78 @@ def show(self) -> None:
     @staticmethod
     def hide(view: sublime.View) -> None:
         mdpopups.hide_popup(view)
+
+
+_phantom_sets_per_view = {}
+
+
+class _PhantonCompletion(BaseCompletion):
```
jfcherng, line 281:
```suggestion
class _PhantomCompletion(BaseCompletion):
```

TheSecEng:
> I've updated the integration a bit, so it now handles multi-line strings and properly ident everything with `&nbsp;` symbols.
> 
> The cursor still jumps in some cases and for now, I am not sure how to fix it(my workaround with an empty block doesn't work in all cases) but in general, it is now a way more reliable

I wonder if we just reset cursor position after drawing the phantom would work.

TheSecEng:
Also, I am assuming we plan to put this behind a setting? So the user can select popup vs phantom?

timfjord:
> > I've updated the integration a bit, so it now handles multi-line strings and properly ident everything with `&nbsp;` symbols.
> > The cursor still jumps in some cases and for now, I am not sure how to fix it(my workaround with an empty block doesn't work in all cases) but in general, it is now a way more reliable
> 
> I wonder if we just reset cursor position after drawing the phantom would work.

I haven't tried this, but the problem with phantom cursor jumping is that the position doesn't really change.  
I think it is basically how Sublime handles phantom rendering.  
I am going to play with it a bit more, but most likely we need to ask someone from the Sublime team (I might try in the discord chat if I won't be able to find a solution).

So as far as I understand, you are also experiencing these problems with the cursor in your implementation?

TheSecEng:
> > I wonder if we just reset cursor position after drawing the phantom would work.
> 
> I haven't tried this, but the problem with phantom cursor jumping is that the position doesn't really change. I think it is basically how Sublime handles phantom rendering. I am going to play with it a bit more, but most likely we need to ask someone from the Sublime team (I might try in the discord chat if I won't be able to find a solution).
> 
> So as far as I understand, you are also experiencing these problems with the cursor in your implementation, right?

Yup, and there is no rhyme or reason to when it happens.

I experienced this and an old plugin as well.

TheSecEng:
For settings, we should really access `session.config.settings` over `view.settings()` as the session holds the most true settings for each LSP session

TheSecEng:
@timfjord Perhaps this is what we experience https://github.com/sublimehq/sublime_text/issues/2871

timfjord:
I think that is exactly it, let me add my comment there.  
A label was added on May'31, so at least it was acknowledged by the Sublime team

TheSecEng:
There is also https://github.com/sublimehq/sublime_text/issues/4575 which might be relevant

timfjord:
image

It was tagged on the same day as the other one, so maybe there is chance it will be implemented

plugin/listeners.py
```diff
@@ -8,10 +8,18 @@
 
 from .plugin import CopilotPlugin
 from .ui import ViewCompletionManager, ViewPanelCompletionManager
-from .utils import get_setting
+from .utils import get_copilot_view_setting, get_setting, set_copilot_view_setting
 
 
 class ViewEventListener(sublime_plugin.ViewEventListener):
+    @property
+    def _is_modifed(self) -> bool:
```
jfcherng, line 16: typo: \_is_modif`i`ed

plugin/ui/completion.py
```diff
@@ -274,3 +295,94 @@ def show(self) -> None:
...
+                line_padding_bottom=int(self.view.settings().get("line_padding_bottom", 0)) * 2,
+            ),
+            sublime.LAYOUT_INLINE if inline else sublime.LAYOUT_BLOCK,
+        )
+
+    def show(self) -> None:
+        completion = self.completion_manager.current_completion
+        assert completion
+
+        head, *tail = completion["displayText"].split("\n")
```
jfcherng, line 373: `str.partition` is perfect for this use. It should be faster and more memory efficient (not important here though 😄 )

timfjord: Unfortunately `str.partition` won't work, because I need tot split text into an array, so I can render each line separately

timfjord:
@jfcherng thanks to your latest debounce improvements the cursor doesn't jump anymore 👍🏻   
I think the main issue with the cursor was that we sent too many requests(because of the old debouncer implementation) and the phantoms were shown and hidden multiple times that was causing the cursor to jump.

There might be still problems with cursor jumping but at least it is way more stable now

plugin/ui/completion.py
jfcherng: > Unfortunately `str.partition` won't work here, because I need to split the text into a list, so I can render each line separately

So you handle the second line in a different way with the third line? I don't expect that. I think only the first and the rest are different.

timfjord: Basically I need a list of lines. 
The first line is rendered as an inline phantom and all other lines are rendered as a block phantom but each line needs to be wrapped in `<div class="copilot-completion-line>`, that's why I need the `split`

jfcherng: image

From my experiment, text in a phantom also auto follows `line_padding_top` and `line_padding_bottom`. There is no need for CSS.

timfjord: It follows only in the inline phantom(first line), and I don't add padding there.  
But, unfortunately, it doesn't follow in the block phantom that is displayed underneath

jfcherng: > But, unfortunately, it doesn't follow in the block phantom that is displayed underneath

ah, I see that now...

jfcherng:
Personally, I have `"line_padding_bottom": 2,` in my preferences.

There is a similar issue to https://github.com/facelessuser/ColorHelper/pull/122.

https://user-images.githubusercontent.com/6594915/182443500-8319948d-228b-4715-8f77-f0fb2a7d1a3c.mp4

plugin/ui/completion.py
```diff
@@ -274,3 +295,94 @@ def show(self) -> None:
...
+
+        self._phantom_set.update(
+            [
+                self._build_phantom(head, completion["point"] + 1, completion["point"]),
```
jfcherng, line 377: Is there a magic? Why the `end` is smaller than `begin`?

plugin/ui/completion.py
```diff
@@ -274,3 +295,94 @@ def show(self) -> None:
...
         mdpopups.hide_popup(view)
+
+
+class _PhantomCompletion(_BaseCompletion):
+    COPILOT_PHANTOM_COMPLETION = "copilot_phantom_completion"
+    PHANTOM_TEMPLATE = """
+    <body id="copilot-completion">
+        <style>
+            body {{
+                color: #808080;
```
jfcherng, line 306: Should we use `comment` color for this? 

E.g.,
```py
>>> view.style_for_scope('comment')
{'bold': False, 'foreground': '#616161', 'italic': False, 'selection_foreground': '#757575', ...}
```

timfjord:
> Personally, I have `"line_padding_bottom": 2,` in my preferences.
> 
> There is a similar issue to [facelessuser/ColorHelper#122](https://github.com/facelessuser/ColorHelper/pull/122).
> 
>  dasf.mp4

An empty block is added to stabilize the cursor and  it causes the current line to be a big bigger.  
I haven't tried `line-height: 0`, but  I will do

plugin/ui/completion.py
timfjord: Yeah, it is a workaround to keep the cursor from the right(unfortunately it doesn't always work, but it is better than nothing)  
There is a note about the empty block, but I will add a note about this trick too

plugin/ui/completion.py
timfjord: I don't know, I took this colour from [the neovim plugin](https://github.com/github/copilot.vim/blob/release/plugin/copilot.vim#L14-L20) and it works ok for both light and dark modes.

What I would do I would add a note on how to customize it, but, IMO, it sounds like a reasonable default

jfcherng: If users don't complain in the neovim plugin, I guess we don't have to do that until many of them complain.

plugin/ui/completion.py
```diff
@@ -274,3 +298,95 @@ def show(self) -> None:
...
+    def show(self) -> None:
+        completion = self.completion_manager.current_completion
+        assert completion
+
+        first_line, *rest_lines = completion["displayText"].splitlines()
+
+        assert self._phantom_set
+        self._phantom_set.update(
+            [
+                self._build_phantom([first_line], completion["point"] + 1, completion["point"]),
```
jfcherng, line 384:
```suggestion
                self._build_phantom(first_line, completion["point"] + 1, completion["point"]),
```

Just notice that I forgot to remove the brackets.

timfjord:
@jfcherng @TheSecEng all the edge cases for the phantom completion are taking longer than planned so I would like to propose the following.

I've addressed the most critical TODOs so to avoid maintaining a long-running branch I would send this one to review.
As for the remaining TODOs, I would separate issues and tackle them separately.
What do you think?

I've also done minor refactoring as part of this PR. 
Basically, I've made the Completion classes more independent and removed `ViewCompletionManager` from its dependencies.

plugin/ui/completion.py
```diff
@@ -84,24 +132,46 @@ def show_next_completion(self) -> None:
         """Show the next completion."""
         self.show(completion_index=self.completion_index + 1)
 
+    def handle_selection_change(self) -> None:
+        if not self.is_phantom() or not self.is_visible:
+            return
+
+        self.hide()
+
+    def handle_text_change(self) -> None:
```
timfjord, line 141: I know that this is is exacrtly the same as the one above, but I am planning to extend it in a follow-up and handle text changes when the phantom completion is visible

timfjord:
Ok, I think it is ready to go

jfcherng (approved):

timfjord:
Hey @TheSecEng 
Have you got a chance to look at this PR?

TheSecEng (approved):

PR #29, opened 2022-07-27
feat: implement getCompletionCycling

TheSecEng:
### What

Fully implement `getCompletionCycling`


plugin/plugin.py
```diff
@@ -171,6 +172,13 @@ def _handle_status_notification_notification(self, payload: CopilotPayloadStatus
         pass
 
     def request_get_completions(self, view: sublime.View) -> None:
+        self._request_completions(view=view, request=REQ_GET_COMPLETIONS)
+        self.request_get_completions_cycling(view=view)
```
jfcherng, line 176: I think this means the result of the first request is always unused but it will make the popup flickering?

jfcherng:
I noticed that I get two completions which are exactly the same.

image
image

jfcherng:
Somehow look at the payload, I feel the debouncing is not working...

TheSecEng:
> I noticed that I get two completions which are exactly the same.
> 
> 
> 
> image
> 
> image
> 
> 

Looking at this

jfcherng:
> Looking at this

Suppose you failed to upload a screenshot?

jfcherng:
> Looking at this

`unique` util may help. https://github.com/TheSecEng/LSP-copilot/blob/1703e6a1351874e23b216a0a664a85182d10476d/plugin/utils.py#L182

jfcherng:
I think the functionality of this PR is working as expect. Let's merge this and move other minor issues to other PRs.

## Potential Issues

- To many requests are sent.
- Flickering completion popups.

timfjord:
I also notice the flickering issue.  
It could be potentially fixed if we won't do another request if there is already a visible completion and the region of the completion is the same as the region of the potential future request. I was going to send a PR later next week

But in general the debouncer requires some research, we might need to also increase the timeout a bit

jfcherng:
> But in general the debouncer requires some research, we might need to also increase the timeout a bit

Yeah. I don't like how that debouncer works at this moment. It just delays requests but never cancels any.

timfjord:
Currently we have `300ms` delay, my idea was to increase it to `600ms` or so. This won't solve the other problems but might at least help stoping some unwanted requests

jfcherng:
> Currently we have `300ms` delay, my idea was to increase it to `600ms` or so. This won't solve the other problems but might at least help stoping some unwanted requests

Even you set it to infinity, it won't stop unwanted requests. It's just defered.

jfcherng:
Try to resolve debounce (and flickering) issue in https://github.com/TheSecEng/LSP-copilot/pull/40

Issue #30, opened 2022-07-27
PHP completions need leading `<?php` to be highlighted properly


jfcherng:
![image](https://user-images.githubusercontent.com/6594915/181248679-a7852125-9ade-4ce8-9ef4-d7ff2da0b137.png)

In the popup, it should use
```php
<?php
$foo = 'bar';
```
to have proper syntax highlighting. This is how LSP-intelephense (PHP language server) shows it with proper highlighting.

PR #32, opened 2022-07-27
fix: PHP syntax highlighting in completions

jfcherng:
image

Fixes https://github.com/TheSecEng/LSP-copilot/issues/30

### DIFF
--- a/plugin/utils.py
+++ b/plugin/utils.py
@@ -65,6 +65,12 @@ def first(items: Iterable[T], test: Optional[Callable[[T], bool]] = None, defaul
     return next(filter(test, items), default)
 
 
+def fix_completion_syntax_highlight(view: sublime.View, point: int, code: str) -> str:
+    if view.match_selector(point, "source.php"):
+        return "<?php\n{}".format(code)
+    return code
+
+
 def get_copilot_view_setting(view: sublime.View, key: str, default: Any = None) -> Any:
     return view.settings().get("{}.{}".format(COPILOT_VIEW_SETTINGS_PREFIX, key), default)
 
--- a/plugin/ui/completion.py
+++ b/plugin/ui/completion.py
@@ -198,12 +198,14 @@ class _PopupCompletion:
     """.format(
         class_name=CSS_CLASS_NAME
     )
+    # We use many backticks to denote a fenced code block because if we are writing in Markdown,
+    # Copilot may suggest 3 backticks for a fenced code block and that can break our templating.
     COMPLETION_TEMPLATE = reformat(
         """
         <div class="header">{header_items}</div>
-        ```{lang}
+        ``````{lang}
         {code}
-        ```
+        ``````
         """
     )
 
--- a/plugin/ui/completion.py
+++ b/plugin/ui/completion.py
@@ -6,7 +6,14 @@ import sublime
 from LSP.plugin.core.typing import List, Optional
 
 from ..types import CopilotPayloadCompletion
-from ..utils import clamp, get_copilot_view_setting, get_view_language_id, reformat, set_copilot_view_setting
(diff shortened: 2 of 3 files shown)


plugin/utils.py
```diff
@@ -65,6 +65,12 @@ def first(items: Iterable[T], test: Optional[Callable[[T], bool]] = None, defaul
     return next(filter(test, items), default)
 
 
+def fix_completion_syntax_highlight(view: sublime.View, point: int, code: str) -> str:
+    if view.match_selector(point, "source.php"):
+        return "<?php\n{}".format(code)
+    return code
```
TheSecEng, line 71: Do we see this being an issue with any other syntax definitions? Or is this unique only to PHP's need for <?php

jfcherng: i don't know. this is one of my familiar language.

jfcherng:
> Can we strip `<?php` after we generate the HTML content?

it's only for display.

plugin/utils.py
jfcherng: The markdown case is more broken than I thought...
image

TheSecEng: image

TheSecEng: Probably this is something that is going to be extremely difficult to detect. We will need to inspect Scope and determine if it is in a code-fence

timfjord: > The markdown case is more broken than I thought... Copilot keeps adding trainling ```.
I think we should escape the complete text.  
For example, this is how I handle this for the phantom completion: https://github.com/TheSecEng/LSP-copilot/blob/e1fada07fc2ce4944b9dcfdfddfc2d5f7c2d8736/plugin/ui/completion.py#L325-L329

Notice `html.escape` part

jfcherng: Is it necessary if we are already in fenced code block?

timfjord: I think it is necessary. Because it treats that ` ``` ` in the completion text as the closing part of the ` ```<lang>` block.

That's why you see ` ``` ` like outside of the highlighted region

jfcherng: > html.escape

how is this revalent to `` ` ``?

and in my experiment, we don't have to escape text which is in fence code block.
image

timfjord: `html.escape` was just an example. In our case we should replace ` ` ` with a Unicode representation(if it even exists)

jfcherng: but from my experiment, there is no need to use `html.escape` for a fenced code block.

I don't think this is what we want.
image

jfcherng: ok great. a simple solution: just use ````` ```` ````` for fenced code block.

timfjord: yeah, we don't need `html.escape`, as `mdpopups` [already covers](https://github.com/facelessuser/sublime-markdown-popups/blob/master/st3/mdpopups/st_code_highlight.py#L88-L114) some basic HTML escaping. Sorry, there was a misunderstanding.

But I would give `code.replace("`", "&#96;")` a go, as it should fix [this](https://github.com/TheSecEng/LSP-copilot/pull/32#discussion_r931196040)

jfcherng: > But I would give `` code.replace("`", "&#96;") `` a go, as it should fix [this](https://github.com/TheSecEng/LSP-copilot/pull/32#discussion_r931196040)

That would break `` ` `` in normal codes such as string literal in JS. A more elegant solution has been found above. :)

TheSecEng:
Fixes https://github.com/TheSecEng/LSP-copilot/issues/33

jfcherng:
> Fixes #33

Nope. #33 is another circumstance.

Issue #33, opened 2022-07-27
Embedded languages are unable to highlight correctly


TheSecEng:
See the below. 

<img width="510" alt="image" src="https://user-images.githubusercontent.com/32599364/181286243-fbebac58-085b-42ff-b5c6-405cd0454fda.png">

Perhaps this is why VSCode doesn't do syntax highlighting, and perhaps we need to turn it off for more complex syntax types like Markdown

PR #34, opened 2022-07-27
fix: wrong syntax highlighting for embedded languages

jfcherng:
Fixes https://github.com/TheSecEng/LSP-copilot/issues/33

---

For example, syntax highlighting for completing JS codes inside Markdown fenced code block.

image

The main idea is to get the deepest/last `source | text | embedding` scope and find the languageId for it.

For the example above, the scope at the cursor is `text.html.markdown markup.raw.code-fence.javascript.markdown-gfm source.js`. We should take `source.js` for resolving languageId rather than `text.html.markdown`.

### DIFF
--- a/plugin/ui/completion.py
+++ b/plugin/ui/completion.py
@@ -215,9 +215,11 @@ class _PopupCompletion:
 
     @property
     def popup_content(self) -> str:
+        self.completion = self.completion_manager.current_completion
+        assert self.completion  # our code flow guarantees this
         return self.COMPLETION_TEMPLATE.format(
             header_items=" &nbsp;".join(self.popup_header_items),
-            lang=get_view_language_id(self.view),
+            lang=get_view_language_id(self.view, self.completion["point"]),
             code=self.popup_code,
         )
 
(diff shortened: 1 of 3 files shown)


TheSecEng (approved):

jfcherng:
Unfortunately, this won't work for PHP.
image

```py
Python 3.3>>> from LSP.plugin.core.types import basescope2languageid
Python 3.3>>> basescope2languageid('source.php.embedded.html')
'html'
```

jfcherng:
I notice that 

- PHP uses `source.php.embedded.html`.
- Markdown uses `source.json.embedded.markdown`. (https://github.com/sublimehq/Packages/blob/17d27049bcc41cb15561476743462c46968fc82b/Markdown/Markdown.sublime-syntax#L317)

So I assume we have a scoping convention for embedded things: `EMBEDDED_LANG.embedded.PARENT_LANG`. Hence we can extract `EMBEDDED_LANG`, which we are interested in.

jfcherng:
Looks good again.

image

Issue #36, opened 2022-07-27
Do we want to maintain `PACKAGE_VERSION` constant manually?


jfcherng:
We send (editor info) plugin version in https://github.com/TheSecEng/LSP-copilot/blob/1703e6a1351874e23b216a0a664a85182d10476d/plugin/plugin.py#L90

If we don't want to maintain that constant manually, we can parse `package-metadata.json` from a installed package and get plugin version from it. As for non-traditional Package Control installation, we may just send unknown.

What do you think?

## package-metadata.json

```js
{
    "dependencies": [],
    "platforms": [
        "*"
    ],
    "sublime_text": ">=3000",
    "version": "0.1.3", // <--------------------------------------------------------------------------
    "url": "https://github.com/alepez/LiveReload-sublimetext3",
    "description": "LiveReload plugin for SublimeText"
}
```

PR #37, opened 2022-07-28
refactor: get plugin version from package metadata file

jfcherng:
Resolves https://github.com/TheSecEng/LSP-copilot/issues/36 if we want this.


TheSecEng:
LGTM

PR #40, opened 2022-07-30
refactor: better completion debounce implementation

jfcherng:
This PR does following things.

- Real debounce implementation.  I.e., only the last call will be executed during a time frame.
- Do not show popup for `REQ_GET_COMPLETIONS` results but only show popup for `REQ_GET_COMPLETIONS_CYCLING` results.
- Remove `CopilotAskCompletionsCyclingCommand` since it makes no sense to use it alone.


timfjord:
It fixes the flickering issue 👍🏻
At least I couldn't reproduce it anymore during my test

timfjord (approved):

PR #46, opened 2022-08-02
tentative refactoring for #27

jfcherng:
No behavior change. Just make codes cleaner imho.

### DIFF
--- a/plugin/ui/completion.py
+++ b/plugin/ui/completion.py
@@ -144,6 +144,8 @@ class _BaseCompletion(metaclass=ABCMeta):
         self.view = view
         self.completion_manager = ViewCompletionManager(view)
 
+        self._settings = self.view.settings()
+
     @abstractmethod
     def show(self) -> None:
         pass
@@ -337,7 +339,7 @@ class _PhantomCompletion(_BaseCompletion):
         return _view_to_phantom_set[view_id]
 
     def normalize_phantom_line(self, line: str) -> str:
-        return html.escape(line).replace(" ", "&nbsp;").replace("\t", "&nbsp;" * self.view.settings().get("tab_size"))
+        return html.escape(line).replace(" ", "&nbsp;").replace("\t", "&nbsp;" * self._settings.get("tab_size"))
 
     def _build_phantom(
         self,
@@ -348,8 +350,6 @@ class _PhantomCompletion(_BaseCompletion):
         inline: bool = True
         # format separator
     ) -> sublime.Phantom:
-        view_settings = self.view.settings()
-
         body = (
             self.normalize_phantom_line(lines)
             if isinstance(lines, str)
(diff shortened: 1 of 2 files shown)


plugin/ui/completion.py
```diff
@@ -323,45 +324,38 @@ class _PhantomCompletion(_BaseCompletion):
...
+        view_settings = self.view.settings()
+
+        body = "".join(
+            self.PHANTOM_LINE_TEMPLATE.format(
+                class_name="rest" if index else "first",
+                content=self.normalize_phantom_line(line),
             )
-            if isinstance(lines, list)
-            else self.normalize_phantom_line(lines)
+            for index, line in enumerate(lines)
```
timfjord, line 350: With this refactoring, `copilot-completion-line` class will be added even for the first inline phantom, but I wanted to avoid this because it adds an extra margin.

plugin/ui/completion.py
```diff
@@ -16,7 +16,8 @@
     set_copilot_view_setting,
 )
 
-_phantom_sets_per_view = {}
+# @todo When a view is closed, it should be deleted from this dict.
```
timfjord, line 19: I've added this as a TODO here https://github.com/TheSecEng/LSP-copilot/pull/27

plugin/ui/completion.py
```diff
@@ -370,19 +364,17 @@ def show(self) -> None:
...
-        head, *tail = completion["displayText"].split("\n")
+        first_line, *rest_lines = completion["displayText"].splitlines()
 
+        assert self._phantom_set
         self._phantom_set.update(
             [
-                self._build_phantom(head, completion["point"] + 1, completion["point"]),
-                # even if the tail is empty it is required to add an empty phantom to prevent the cursor from jumping
-                self._build_phantom(tail or "", completion["point"], inline=False),
+                self._build_phantom([first_line], sublime.Region(completion["point"] + 1, completion["point"])),
```
timfjord, line 372: I personally liked the previous implementation with points more, mostly because it was shorter

jfcherng: I will revise that part back later.

plugin/ui/completion.py
jfcherng: Oops, didn't notice this. I think there is no good common way to refactor this then. Will revert.

plugin/ui/completion.py
```diff
@@ -323,45 +324,50 @@ class _PhantomCompletion(_BaseCompletion):
...
-        self, lines: Union[str, List[str]], begin: int, end: Optional[int] = None, inline: bool = True
+        self,
+        lines: Union[str, Sequence[str]],
+        begin: int,
+        end: Optional[int] = None,
+        *,
+        inline: bool = True
+        # format separator
     ) -> sublime.Phantom:
+        view_settings = self.view.settings()
```
timfjord, line 351: Just thinking out loud, is it worth adding `self._settings = self.view.settings()` in the `__init__` ?

We also use view settings in the `normalize_phantom_line` method

jfcherng: done.

timfjord (approved):

PR #50, opened 2022-08-15
tentative refactoring for #27

jfcherng:
Besides, I don't like the following. `ViewCompletionManager` shouldn't be coupled with `phantom`-related thing.

```py
class ViewCompletionManager:
    @property
    def is_phantom(self) -> bool:
        return self.completion_style == _PhantomCompletion.STYLE_TYPE
        
    def handle_selection_change(self) -> None:
        if not (self.is_phantom and self.is_visible):
            return

        self.hide()

    def handle_text_change(self) -> None:
        if not (self.is_phantom and self.is_visible):
            return

        self.hide()

    def handle_close(self) -> None:
        if not self.is_phantom:
            return

        self.completion_style_type.close(self.view)
```

### DIFF
--- a/plugin/ui/completion.py
+++ b/plugin/ui/completion.py
@@ -175,7 +175,7 @@ class ViewCompletionManager:
 
 
 class _BaseCompletion(metaclass=ABCMeta):
-    STYLE_TYPE = ""
+    name = ""
 
     def __init__(
         self,
--- a/plugin/ui/completion.py
+++ b/plugin/ui/completion.py
@@ -202,7 +202,7 @@ class _BaseCompletion(metaclass=ABCMeta):
 
 
 class _PopupCompletion(_BaseCompletion):
-    STYLE_TYPE = "popup"
+    name = "popup"
 
     CSS_CLASS_NAME = "copilot-completion-popup"
     CSS = """
--- a/plugin/ui/completion.py
+++ b/plugin/ui/completion.py
@@ -343,7 +343,7 @@ class _PopupCompletion(_BaseCompletion):
 
 
 class _PhantomCompletion(_BaseCompletion):
-    STYLE_TYPE = "phantom"
+    name = "phantom"
 
     COPILOT_PHANTOM_COMPLETION = "copilot_phantom_completion"
     PHANTOM_TEMPLATE = """
--- a/plugin/ui/completion.py
+++ b/plugin/ui/completion.py
@@ -92,7 +92,7 @@ class ViewCompletionManager:
 
     @property
     def completion_style_type(self) -> Type["_BaseCompletion"]:
-        completion_cls = first(_BaseCompletion.__subclasses__(), lambda t: t.STYLE_TYPE == self.completion_style)
+        completion_cls = first(_BaseCompletion.__subclasses__(), lambda t: t.name == self.completion_style)
         if completion_cls:
             return completion_cls
         raise RuntimeError("Unknown completion style type: {}".format(self.completion_style))
--- a/plugin/ui/completion.py
+++ b/plugin/ui/completion.py
@@ -9,6 +9,7 @@ from LSP.plugin.core.typing import Dict, List, Optional, Sequence, Type, Union
 from ..types import CopilotPayloadCompletion
 from ..utils import (
     clamp,
+    first,
     fix_completion_syntax_highlight,
     get_copilot_view_setting,
     get_view_language_id,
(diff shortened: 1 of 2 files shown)


plugin/ui/completion.py
```diff
@@ -78,15 +55,11 @@ def completions(self, value: List[CopilotPayloadCompletion]) -> None:
     @property
     def completion_style(self) -> str:
         """The completion style."""
-        return get_copilot_view_setting(self.view, "completion_style", self.COMPLETION_STYLES[0])
+        return get_copilot_view_setting(self.view, "completion_style", "")
```
jfcherng, line 58: If it will break, then let it break (exception in `completion_style_type`). The user should fix it.

plugin/ui/completion.py
```diff
@@ -118,11 +91,15 @@ def current_completion(self) -> Optional[CopilotPayloadCompletion]:
         return self.completions[self.completion_index] if self.completions else None
 
     @property
-    def completion_style_type(self) -> Type[_BaseCompletion]:
-        return globals().get("_{}Completion".format(self.completion_style.capitalize()))
+    def completion_style_type(self) -> Type["_BaseCompletion"]:
```
timfjord, line 94: Strange, `pyright` complained when I tried `Type["_BaseCompletion"]`

plugin/ui/completion.py
```diff
@@ -118,11 +91,15 @@ def current_completion(self) -> Optional[CopilotPayloadCompletion]:
...
+        completion_cls = first(_BaseCompletion.__subclasses__(), lambda t: t.STYLE_TYPE == self.completion_style)
```
timfjord, line 95:
```suggestion
        completion_cls = first(_BaseCompletion.__subclasses__(), lambda t: t.name == self.completion_style)
```

plugin/ui/completion.py
```diff
@@ -118,11 +91,15 @@ def current_completion(self) -> Optional[CopilotPayloadCompletion]:
...
+        if completion_cls:
+            return completion_cls
+        raise RuntimeError("Unknown completion style type: {}".format(self.completion_style))
 
+    @property
     def is_phantom(self) -> bool:
-        return self.completion_style == self.COMPLETION_STYLE_PHANTOM
+        return self.completion_style == _PhantomCompletion.STYLE_TYPE
```
timfjord, line 102:
```suggestion
        return self.completion_style == _PhantomCompletion.name
```

plugin/ui/completion.py
```diff
@@ -200,7 +174,36 @@ def _tidy_completion_index(self, index: int) -> int:
         return clamp(index, 0, completions_cnt - 1)
 
 
+class _BaseCompletion(metaclass=ABCMeta):
+    STYLE_TYPE = ""
```
timfjord, line 178:
```suggestion
    name = ""
```

plugin/ui/completion.py
```diff
@@ -340,6 +343,8 @@ def hide(view: sublime.View) -> None:
 
 
 class _PhantomCompletion(_BaseCompletion):
+    STYLE_TYPE = "phantom"
```
timfjord, line 346:
```suggestion
    name = "phantom"
```

plugin/ui/completion.py
```diff
@@ -200,7 +174,36 @@ def _tidy_completion_index(self, index: int) -> int:
...
+        pass
+
+    @classmethod
+    @abstractmethod
+    def hide(cls, view: sublime.View) -> None:
+        pass
+
+
 class _PopupCompletion(_BaseCompletion):
+    STYLE_TYPE = "popup"
```
timfjord, line 205:
```suggestion
    name = "popup"
```

timfjord (commented):
I like the idea of this refactoring, but I think we should not use constants here, as the  `STYLE_TYPE` is not a constant, because we override it in child classes.

I also think we need to use a more generic name(e.g. `PhantomCompletion.name`)

As for the other concern, regarding coupling, theoretically, we could just proxy all those methods to the `completion_style_type` and let it handle everything, but at the same time, we still need to know whether the completion is visible or not. Maybe we could go hybrid, and do some check in the manager and other checks in the completion class itself

plugin/ui/completion.py
jfcherng: Maybe it's due to `globals().get(...)`, which too dynamic so pyright can't determinate whether it's `_BaseCompletion`?

timfjord: Probably

timfjord (approved):

PR #57, opened 2022-08-18
chore: use mypy for CI

jfcherng:
Using submodule is probably easier than maintaining stubs?


timfjord:
Maybe we could use https://github.com/github/super-linter?

jfcherng:
Never tried it but still I prefer having the ability to do it local (in our case, the Makefile).

PR #59, opened 2022-09-19
fix: do not get completions during the lsp_save command

timfjord:
Some LSP servers are slower than others, and, for example, formatting on save (that is caused by the `lsp_save` command) can take a bit longer.  
Because of that, the `on_modified_async` callback gets triggered in the middle of the call and a completion gets displayed(because it may have formatted the source code).

This PR blocks the completion request during the `lsp_save` command duration

### DIFF
--- a/plugin/listeners.py
+++ b/plugin/listeners.py
@@ -16,6 +16,14 @@ class ViewEventListener(sublime_plugin.ViewEventListener):
     def _is_modified(self, value: bool) -> None:
         set_copilot_view_setting(self.view, "_is_modified", value)
 
+    @property
+    def _is_saving(self) -> bool:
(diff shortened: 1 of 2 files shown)


timfjord:
@TheSecEng @jfcherng any comments on this one?

jfcherng:
I am not sure how much this improves UX since I don't use `lsp_save`. Though it can take a long time but I presume `lsp_save` won't set the view to be readonly so that user can still modify the view per se during save. So what will happen if the user modify the buffer during, say, a formatting which takes 10 seconds? 

If it's allowed to modify the buffer during formatting, then I think it makes sense to request completions.

jfcherng (approved):

TheSecEng:
LGTM

Issue #52, opened 2022-08-17
Handle text changes when phantom completion is visible


timfjord:
Currently, when a phantom completion is displayed typing characters that match the beginning of the ghost text hides the completion and instantiates a new request.

This can be improved and the currently displayed ghost text can be updated without triggering a new request when typed characters match the begging of the ghost text

TheSecEng:
I wonder if this could interfere with copilot returning more accurate results. Or different results. 

Perhaps we should update the phantom and still request. And if the result doesn't meet the current completion we update.

timfjord:
As far as I understand, in VSCode they don't do any requests.  
At least the status indicator doesn't show anything when you type characters that match the beginning of the ghost text, but I think it is handled on the core level because they use the suggestion API(or what is the official name for it) and it works the same for other completions.

We can do some requests, but I am afraid the experience won't be smooth.  
Plus, I don't think that we would see different results there

I would handle this in a follow-up maybe

PR #60, opened 2022-10-04
fix: align accept completion logic with VSCode

timfjord:
In VSCode when the cursor is at the beginning of the line and there is a visible completion, the completion gets accepted only when the cursor is literally near the completion.  
So, for example, when the cursor is at the beginning of the line and the completion is a few tabs/stop away then accepting a completion(e.g. pressing the tab) won't accept the completion and instead trigger the tab key.

https://user-images.githubusercontent.com/471335/193867633-2130301a-4591-466a-8e80-8e350f3321e4.mov

This PR emulates this logic by adding a query context that is used by default.  
It is important to either keep this logic default or remove the default keymap for the tab at all because otherwise there will be no way to override the tab key in the user keymaps.

https://user-images.githubusercontent.com/471335/193867564-73603493-8ee6-48e4-9d47-64ea10c64714.mov

Also, this change should play nicely with this one https://github.com/TheSecEng/LSP-copilot/issues/52  
So once https://github.com/TheSecEng/LSP-copilot/issues/52 is implemented the completion will be updated right away without an extra request/delay(similar to VSCode)

### DIFF
--- a/plugin/listeners.py
+++ b/plugin/listeners.py
@@ -66,10 +66,7 @@ class ViewEventListener(sublime_plugin.ViewEventListener):
         if key == "copilot.is_on_completion":
             vcm = ViewCompletionManager(self.view)
 
-            if not vcm.is_visible:
-                return test(False)
-
-            if len(self.view.sel()) < 1:
+            if not vcm.is_visible or len(self.view.sel()) < 1 or not vcm.current_completion:
                 return test(False)
 
             point = self.view.sel()[0].begin()
--- a/plugin/listeners.py
+++ b/plugin/listeners.py
@@ -73,9 +73,7 @@ class ViewEventListener(sublime_plugin.ViewEventListener):
             line = self.view.line(point)
             beginning_of_line = self.view.substr(sublime.Region(line.begin(), point))
 
-            return test(
-                beginning_of_line.strip() or re.match(r"(?!\s)", str(vcm.current_completion["displayText"]))
-            )
+            return test(beginning_of_line.strip() or not re.match(r"\s", vcm.current_completion["displayText"]))
 
         return None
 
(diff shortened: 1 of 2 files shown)


plugin/listeners.py
```diff
@@ -57,8 +59,27 @@ def test(value: Any) -> Optional[bool]:
...
+
+            if len(self.view.sel()) < 1:
+                return test(False)
+
+            point = self.view.sel()[0].begin()
+            line = self.view.line(point)
+            beginning_of_line = self.view.substr(sublime.Region(line.begin(), point))
+
+            return test(
+                beginning_of_line.strip() != "" or not re.match(r"\s+", str(vcm.current_completion["displayText"]))
```
jfcherng, line 80:
```suggestion
                beginning_of_line.strip() or re.match(r"(?!\s)", str(vcm.current_completion["displayText"]))
```

plugin/listeners.py
```diff
@@ -57,8 +59,27 @@ def test(value: Any) -> Optional[bool]:
...
             return test(CopilotPlugin.get_account_status().is_authorized)
+
+        if key == "copilot.is_on_completion":
+            vcm = ViewCompletionManager(self.view)
+
+            if not vcm.is_visible:
+                return test(False)
+
+            if len(self.view.sel()) < 1:
+                return test(False)
```
jfcherng, line 73:
```suggestion
            if not vcm.is_visible or len(self.view.sel()) < 1 or not vcm.current_completion:
                return test(False)
```

jfcherng:
Looks good. Just need to run `black`.

timfjord:
Great.
Some of the suggestions have broken the functionality. Checking everything now

timfjord:
Great, now it is working. The regex caused the issue

jfcherng (approved):

timfjord:
Just found another issue with the implementation

Issue #54, opened 2022-08-17
Cursor jumps when completion_style is phantom


timfjord:

timfjord:
This has to do with the way how Sublime Text renders phantoms.  
Currently, there is no way to ensure a phantom will be rendered from the right - https://github.com/sublimehq/sublime_text/issues/4575

We use a workaround, but, unfortunately, it doesn't work all the time

PR #62, opened 2022-10-05
feat: update the phantom while typing

### LINKED ISSUE #52: Handle text changes when phantom completion is visible

timfjord:
Closes #52

This PR tries to emulate VSCode when the phantom(ghost) text is updated while typing(without triggering an extra request).  
Unfortunately, due to the phantom API limitations, it is currently not possible(see the attached video)

https://user-images.githubusercontent.com/471335/194121909-c98ed8eb-4ae7-4221-ac32-b7bd7a1b7cfa.mov

The main problem is that the cursor always jumps when updating the phantom in real time.

### DIFF
--- a/plugin/listeners.py
+++ b/plugin/listeners.py
@@ -34,7 +34,8 @@ class ViewEventListener(sublime_plugin.ViewEventListener):
             return
 
         vcm = ViewCompletionManager(self.view)
-        vcm.handle_text_change()
+        if vcm.handle_text_change():
+            return
 
         if not self._is_saving and get_session_setting(session, "auto_ask_completions") and not vcm.is_waiting:
             plugin.request_get_completions(self.view)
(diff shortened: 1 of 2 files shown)


timfjord:
I opened it just to have some reference but, in general, this PR should be merged only when the problem with the cursor is fixed(see https://github.com/sublimehq/sublime_text/issues/5536 and https://github.com/TheSecEng/LSP-copilot/issues/54)

jfcherng:
I feel the UX in the video is worse.

timfjord:
Yes, it is horrible. That's why we shouldn't merge this one.

We could close it, as I opened it purely to have some reference

jfcherng:
> We could close it, as I opened it purely to have some reference

Keep it open is fine to make use remember this 😄

PR #66, opened 2022-11-07
docs: add FAQ for working under Zscaler environment

jfcherng:
As the conclusion of https://github.com/TerminalFi/LSP-copilot/issues/65


TerminalFi:
I'm not sure this needs to be Zscalar specific. As you'll run into this with many other proxies. Probably best to write a general proxy section.

jfcherng:
I am not sure whether this is a Zscaler issue or cert issue since I don't use any of them.

TerminalFi (approved):

PR #67, opened 2022-11-09
chore: bump server to 1.6.1

jfcherng:
as the title


jfcherng:
To be clear, I can't push to the master branch since it requires 1 approval now.

TerminalFi (approved):

PR #69, opened 2022-11-27
refactor: remove deprecated method

jfcherng:
It has been more than a month since last `lsp_utils` release so I guess it's safe to remove deprecated `minimum_node_version` (it has been replaced by `required_node_version`).


jfcherng:
The test fail looks like a false positive to me. They are used with `# type: ...`. But sure, technically it's "unused" in grammatic view...

timfjord (approved):

PR #79, opened 2023-01-20
fix: Invalid params: /doc must have required property 'version'

TerminalFi:
Copilot now requests that the `doc` parameter now contains `version` with the document version. This must match what is on the server side as well. 

For now, using value of 0 seems to work, however the best scenario would be to try and get the `version` from LSP as it tracks it for us

### DIFF
--- a/language-server/package-lock.json
+++ b/language-server/package-lock.json
@@ -6,20 +6,20 @@
     "packages": {
         "": {
             "dependencies": {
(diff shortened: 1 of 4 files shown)


timfjord (approved):

jfcherng:
Confirmed working.

jfcherng (approved):

jfcherng:
I can help fixing that linting issue.

PR #95, opened 2023-03-16
feat: add HTTP proxy support

jfcherng:
Resolves https://github.com/TerminalFi/LSP-copilot/issues/94

Add a `proxy` setting to `settings` in `LSP-copilot.sublime-settings`.

```js
{
	"settings": {
		"proxy": "username:password@host:port" // or just "host:port"
	}
}
```

### DIFF
--- a/plugin/plugin.py
+++ b/plugin/plugin.py
@@ -3,10 +3,12 @@ import json
 import os
 import weakref
 from functools import wraps
+from urllib.parse import urlparse
 
 import sublime
 from LSP.plugin import Request, Session
-from LSP.plugin.core.typing import Any, Callable, Optional, Tuple, Union, cast
+from LSP.plugin.core.collections import DottedDict
+from LSP.plugin.core.typing import Any, Callable, Dict, Optional, Tuple, Union, cast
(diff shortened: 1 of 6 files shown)


timfjord (approved):

plugin/plugin.py
```diff
@@ -117,20 +120,33 @@ def on_set_editor_info(result: str, failed: bool) -> None:
...
+                "rejectUnauthorized": True,
+            }
+
+        session = self.weaksession()
+        if not session:
+            return
+
+        editor_info = self.editor_info()
+
+        networkProxy = parse_proxy(settings.get("proxy") or "")
```
jfcherng, line 145: Oops... should use snake_cased `network_proxy` as the variable name.

pramodhrachuri:
Any plans on adding socks5 proxy as well?
SSH supports tunneling using socks5. So, might be useful.
(Personally, I use it a lot haha)

jfcherng:
> Any plans on adding socks5 proxy as well? SSH supports tunneling using socks5. So, might be useful. (Personally, I use it a lot haha)

We don't implement actual feature here. The key is the server support it and we just sent an option to it. If you are interested in socks5, you can follow https://github.com/orgs/community/discussions/29127.

PR #100, opened 2023-03-30
fix: completions panel steals focus

nsubiron:
Hi there!

I've been trying to use the completions panel with key bindings only, but right now it's not possible because the panel steals focus from the view. This prevents any further view commands and forces you to close the panel using the mouse.

This PR just gives focus back to the view after opening the panel :)

### DIFF
--- a/plugin/ui/panel_completion.py
+++ b/plugin/ui/panel_completion.py
@@ -279,6 +279,8 @@ class _PanelCompletion:
         else:
             self._open_in_group(window, active_group + 1)
 
+        window.focus_view(self.view)
+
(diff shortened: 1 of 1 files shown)


plugin/ui/panel_completion.py
```diff
@@ -279,6 +279,8 @@ def open(self) -> None:
         else:
             self._open_in_group(window, active_group + 1)
 
+        window.focus_view(self.view)
```
TerminalFi, line 282: Seems simple enough. Just don't want to approve in a silo.

TerminalFi (approved):

PR #103, opened 2023-04-19
fix: popup doesn't work properly for cloned views

### LINKED ISSUE #102: No Copilot suggestions if same file is opened in a separate pane (i.e. Sublime "column")?
See description and repro video below :-)

When the same file is opened in separate pane (I know that it feels weird -- but sometimes I forget that a file is a tab in one pane, and I open it again in a different pane), Copilot suggestions don't appear. It seems that suggestions only appear if the file is opened once.

https://user-images.githubusercontent.com/1017304/233140718-5cca3d99-3fc4-4372-b76a-a3ac2cb3a6ff.mov

jfcherng:
fixes #102

### DIFF
--- a/plugin/listeners.py
+++ b/plugin/listeners.py
@@ -1,15 +1,37 @@
 import re
+from functools import wraps
 
 import sublime
 import sublime_plugin
-from LSP.plugin.core.typing import Any, Dict, Optional, Tuple
+from LSP.plugin.core.typing import Any, Callable, Dict, Optional, Tuple, cast
 
 from .plugin import CopilotPlugin
+from .types import T_Callable
 from .ui import ViewCompletionManager, ViewPanelCompletionManager
-from .utils import get_copilot_view_setting, get_session_setting, set_copilot_view_setting
+from .utils import get_copilot_view_setting, get_session_setting, is_activate_view, set_copilot_view_setting
+
+
+def _must_be_active_view(*, failed_return: Any = None) -> Callable[[T_Callable], T_Callable]:
+    def decorator(func: T_Callable) -> T_Callable:
+        @wraps(func)
+        def wrapped(self: Any, *arg, **kwargs) -> Any:
+            if is_activate_view(self.view):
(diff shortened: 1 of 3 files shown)


plugin/utils.py
```diff
@@ -88,6 +88,10 @@ def find_view_by_id(id: int) -> Optional[sublime.View]:
     return first(all_views(include_transient=True), lambda view: view.id() == id)
 
 
+def is_activate_view(obj: Any) -> bool:
```
rchl, line 91: `activate` or `active`?

plugin/listeners.py
```diff
@@ -26,6 +48,7 @@ def _is_saving(self) -> bool:
     def _is_saving(self, value: bool) -> None:
         set_copilot_view_setting(self.view, "_is_saving", value)
 
+    @_must_be_active_view()
```
rchl, line 51: Would it work to use `self.view.is_primary()`?

Checking if the view is active sounds rather brittle. What if view is not active? You'd potentially miss some update?

jfcherng: If we use `self.view.is_primary()`, then when the user is typing in a non-primary view, how to make the popup appear in the non-primary view?

rchl: If it's just about whether to show the popup then I guess it's fine.
I'm just thinking that the document can get modified when in the background but maybe that's not a problem here.

jfcherng: > I'm just thinking that the document can get modified when in the background but maybe that's not a problem here.

Yes. If it's modified in background, I think it makes no sense to fire an Copilot request for it.

TerminalFi (approved):

timfjord (approved):

plugin/listeners.py
TerminalFi: @jfcherng I've been using this for awhile now with no foreseeable issues. It may be improved upon however at the moment I think this is working in a better respect that in the current release

PR #104, opened 2023-04-30
adding version info into status bar

TerminalFi:
image

### DIFF
--- a/plugin/plugin.py
+++ b/plugin/plugin.py
@@ -20,11 +20,13 @@ from .constants import (
     REQ_CHECK_STATUS,
     REQ_GET_COMPLETIONS,
     REQ_GET_COMPLETIONS_CYCLING,
+    REQ_GET_VERSION,
     REQ_SET_EDITOR_INFO,
 )
 from .types import (
     AccountStatus,
     CopilotPayloadCompletions,
+    CopilotPayloadGetVersion,
     CopilotPayloadLogMessage,
     CopilotPayloadPanelSolution,
     CopilotPayloadSignInConfirm,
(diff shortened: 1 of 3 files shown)


plugin/plugin.py
```diff
@@ -119,8 +121,16 @@ def on_check_status(result: CopilotPayloadSignInConfirm, failed: bool) -> None:
         def on_set_editor_info(result: str, failed: bool) -> None:
             pass
 
+        def _set_version_info(response: CopilotPayloadGetVersion, failed: bool) -> None:
+            version = response.get("version")
+            if version:
+                session = self.weaksession()
+                if session:
+                    session.set_config_status_async(version)
```
rchl, line 129: Does it make sense for the user to know the version? This takes extra space so if it's not relevant to all or most users then it's just noise and wasted space.

The intention of being able to dynamically set the status was for example to show the version in cases where the version is dynamic depending on the project (like in case of typescript) where it's beneficial to see at a glimpse what version is used by the project. If the version is static and just depends on version of the ST package then IMO it's not really that useful.

Or if you insist on having it then it could also be something that could be toggled.

I don't say that as a user of this package so I don't really care personally. I just don't think that there should be a version shown just because it's possible.

TerminalFi: I’ve found it useful mainly because I work in a multi version environment where some of them are mainly set to a binary and some are managed by the package. 

That said, I do see that it takes up status bar space. So perhaps I can just create a Check Version command and have it under in the command palette and then let users add an entry if they need it. 

I’ll redirect this PR to do that instead as it fulfills my purposes

rchl: And note that especially in the case of this server, the version seems hardly relevant for a typical user since I can't even find some public changelogs that the user could reference to see what is the difference between 1.73.0 and 1.70.0, for example.

I'm not even sure if there is any benefit in knowing that for you, as a developer of this package. :)

PR #106, opened 2023-05-05
Supports the ability to honor `.gitignore` patterns (TOP LEVEL ONLY)

TerminalFi:
Attempt to have LSP-copilot honor `.gitignore` files that are detected at the top level of the window folders. Comparing the View path with the window path to get the correct ignore file.

### DIFF
--- a/LSP-copilot.sublime-settings
+++ b/LSP-copilot.sublime-settings
@@ -16,7 +16,8 @@
 		"local_checks": false,
 		"telemetry": false,
 		"proxy": "",
-		"completion_style": "popup"
+		"completion_style": "popup",
+		"respect_gitignore": false
 	},
 	// ST4 configuration
 	"selector": "source | text | embedding"
--- a/plugin/listeners.py
+++ b/plugin/listeners.py
@@ -8,7 +8,13 @@ from LSP.plugin.core.typing import Any, Callable, Dict, Optional, Tuple, cast
 from .plugin import CopilotPlugin
 from .types import T_Callable
 from .ui import ViewCompletionManager, ViewPanelCompletionManager
-from .utils import get_copilot_view_setting, get_session_setting, is_active_view, set_copilot_view_setting
+from .utils import (
+    get_copilot_view_setting,
+    get_session_setting,
+    is_active_view,
+    is_gitignored,
+    set_copilot_view_setting,
+)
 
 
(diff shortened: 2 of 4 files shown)


jfcherng:
> `.gitignore` files that are detected at the top level of the window folders.

Arguably this is not how `.gitignore` works.

TerminalFi:
> > `.gitignore` files that are detected at the top level of the window folders.
> 
> Arguably this is not how `.gitignore` works. It should be searched bottom-up until `.git` is seen even if it's not in the workspace. (what if there is `.gitignore` but there is no `.git`? 😂)
> 
> Ok. I see **TOP LEVEL ONLY**.

Yes this is a very basic implementation, I have a more complete one in the works. But I also am expecting copilot to release a spec for this so not sure how much effort to put into this. 

At the moment I can push more which does searching bottom up until it reaches the workspace. Those changes will land soon.

jfcherng:
> Yes this is a very basic implementation, I have a more complete one in the works. But I also am expecting copilot to release a spec for this so not sure how much effort to put into this.

Yes, making a feature request to Copilot is an option too.

rchl:
Also it's not correct to treat all entries in .gitignore as paths relative to its directory. If the entry doesn't start with a slash then it applies also in subdirectories. and there are probably more rules, i haven't read the spec.

TerminalFi:
> Also it's not correct to treat all entries in .gitignore as paths relative to its directory. If the entry doesn't start with a slash then it applies also in subdirectories. and there are probably more rules, i haven't read the spec.

Yup, I’ve got this solved as well I believe

LSP-copilot.sublime-settings
```diff
@@ -16,7 +16,8 @@
 		"local_checks": false,
 		"telemetry": false,
 		"proxy": "",
-		"completion_style": "popup"
+		"completion_style": "popup",
+		"respect_gitignore": false
```
timfjord, line 20: I would add a comment here and in the `sublime-package.json`'s `markdownDescription` section that the lookup is top level only

timfjord (approved):

LSP-copilot.sublime-settings
TerminalFi: I'll be pushing a change shortly that should have it functioning pretty closely to gitignore spec

TerminalFi:
Copilot will natively support it soon.

PR #116, opened 2023-08-04
chore: bump server to 1.9.2

jfcherng:
Tested working on my Win11 PC.

### DIFF
--- a/plugin/plugin.py
+++ b/plugin/plugin.py
@@ -85,9 +85,8 @@ class CopilotPlugin(NpmClientHandler):
         server_directory,
         "node_modules",
         "copilot-node-server",
-        "copilot",
-        "dist",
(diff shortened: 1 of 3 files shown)


plugin/plugin.py
```diff
@@ -85,9 +85,8 @@ class CopilotPlugin(NpmClientHandler):
         server_directory,
         "node_modules",
         "copilot-node-server",
-        "copilot",
-        "dist",
-        "agent.js",
+        "bin",
+        "copilot-node-server",
```
jfcherng, line 89: Using a fixed path binary is better. Introduced in https://github.com/jfcherng/copilot-node-server/pull/2/files

TerminalFi (approved):

PR #124, opened 2023-09-30
Added Settings section to README.md

morehawes:
Thanks for LSP-copilot! To understand the settings I searched and came across [this issue](https://github.com/TerminalFi/LSP-copilot/issues/110) which was helpful. I think including this on the repo README.md makes sense.

Added section/table to README.md describing the Type/Default/Description for the available Settings as described in the sublime-package.json. Hopefully others will find this helpful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ On the first time use, follow the steps below:
 
 ## Settings
 
-Provide user settings in LSP-copilot.sublime-settings (Settings... > Package Settings > LSP > Servers > LSP-copilot)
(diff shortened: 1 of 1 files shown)


TerminalFi (changes requested):
Instead of Settngs > Lsp ETC. 

Update to say they can type `Preferences: LSP-copilot Settings` in the command palette

morehawes:
Good call, I have changed that and tweaked the wording slightly.

PR #136, opened 2024-03-03
chore: bump server to 1.24.0

jfcherng:
Hmm... somehow `node bin/copilot-node-server --stdio` doesn't work while `node copilot/dist/agent.js --stdio` works. Temporarily switch to `copilot/dist/agent.js`.


jfcherng:
CI failures are not important. It looks like due to `black` has change their formatting or something.

TerminalFi (approved):

PR #139, opened 2024-03-13
chore: change copilot_accept_completion keybinding to ctrl+tab

### LINKED ISSUE #73: Copilot completes on tab despite rebinding copilot_accept_completion
Thanks for working on this package, and bringing copilot to Sublime Text!

In my `Default.sublime-keymap` I have bound `copilot_accept_completion` like this:

```json
  {
    "keys": ["primary+enter"],
    "command": "copilot_accept_completion",
    "context": [
      {
        "key": "setting.copilot.completion.is_visible"
      }
    ]
  },
```

Despite that, the default key binding using `"tab"` still exists and takes precedence over other LSP autocompletion.

How do I disable that? Sublime Text now has 4 other bindings for tab, do I have to copy all of them into my key bindings (b/c *that* seems to work)? That’s inconvenient and might break functionality introduced in the future.

### LINKED ISSUE #138: Completing when I press tab, even though hook_to_auto_complete_command is false
After restarting yesterday and presumably having my ST packages updated, LSP-copilot has been auto-inserting suggested text when I press tab (my auto_complete trigger).

I am using the default `"hook_to_auto_complete_command": false,` settings.

jfcherng:
# This change will break almost ALL LSP-copilot user's setup.

`ctrl+tab` is just an example because I don't know what would be suitable for this.

Previously it uses Tab, which is the same with ST's autocompletion, as the keybinding. It's not obvious that what should happen after pressing Tab if there are both copilot/ST completions on the screen. One user may want to use copilot's while another may want ST's.

Or, we assign no keybinding and let users to do their own. (Need to be documented in README.)

---

Resolves #138
Resolves #73

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -1,7 +1,7 @@
 [
     {
         "keys": [
-            "tab"
+            "ctrl+tab"
         ],
         "command": "copilot_accept_completion",
         "context": [


jfcherng:
Ideally I would like some people respond to this. Maybe there are other solutions which I didn't figure out.

rchl:
You can't use ctrl+tab since that's used for switching tabs.
Also with such changes you would also want to add a changelog that highlights the change.

timfjord:
Rather than introducing a new keybinding, that can lead to some issues with existing users as well as with those who rely on `ctrl+tab`, we could handle this with an additional context item.
We could add a setting, say, `copilot_tab_completion`(or something else) to the package settings and make it `true` by default.
Then add a context item to the `tab` keybinding
```jsonc
{
    "keys": [
        "tab"
    ],
    "command": "copilot_accept_completion",
    "context": [
        { "key": "copilot.is_on_completion" },
        { "key": "setting.copilot_tab_completion" },
    ]
}
```
So users who don't want the default behaviour will just set it to `false` and add whatever keybinding they like.

jfcherng:
> Rather than introducing a new keybinding, that can lead to some issues with existing users as well as with those who rely on `ctrl+tab`, we could handle this with an additional context item. We could add a setting, say, `copilot_tab_completion`(or something else) to the package settings and make it `true` by default. Then add a context item to the `tab` keybinding
> 
> ```js
> {
>     "keys": [
>         "tab"
>     ],
>     "command": "copilot_accept_completion",
>     "context": [
>         { "key": "setting.copilot_tab_completion" },
>         { "key": "copilot.is_on_completion" }
>     ]
> }
> ```
> 
> So users who don't want the default behaviour will just set it to `false` and add whatever keybinding they like.

This sounds decent!

jfcherng:
Closed as https://github.com/TerminalFi/LSP-copilot/pull/139#issuecomment-1993830542 is obviously better.

timfjord:
I am just not 100% sure Sublime allows custom packages to be references with `setting.*` syntax
But we can always add a custom context(e.g. `copilot.tab_tab_completion`)

PR #157, opened 2024-05-29
fix: unauthorized account failed to sign out

jfcherng:
Fixes https://github.com/TerminalFi/LSP-copilot/issues/156


TerminalFi:
Before merging, I'd like to look into this a bit.

TerminalFi:
@jfcherng This is odd, but the command works for me on MacOS?

jfcherng:
> @jfcherng This is odd, but the command works for me on MacOS?

Even after reloggin?

jfcherng:
Some observation, 

- previously, the copilot's session file is `hosts.json` but now it's `apps.json`.
- my first sign out success (I believe at that moment it uses `hosts.json`)
- I re-sign in. After that, I failed to sign out.

TerminalFi:
Mine just switched over to apps.json and now signout is not working. Let me dig some more. I would image, it is the agent.js that is failing here. Since it should be the one handling the actual signout logic

jfcherng:
> Mine just switched over to apps.json and now signout is not working. Let me dig some more. I would image, it is the agent.js that is failing here. Since it should be the one handling the actual signout logic

Yes, but it's weird that Copilot.vim seems just calls the `signOut` command and that's all. [github/copilot.vim@c79d711/autoload/copilot.vim#L630-L638](https://github.com/github/copilot.vim/blob/c79d711cbf7c6672c6c57d6df7c5ab7b6cac2b7a/autoload/copilot.vim#L630-L638)

But since it intentionally disable the issue page on GitHub, I guess my report may not be seen by staff.

PR #168, opened 2024-06-19
🎊  Copilot Chat

TerminalFi:
Current status

TerminalFi 2024-07-02 at 15 29 07@2x


plugin/client.py
```diff
@@ -197,7 +198,7 @@ def version() -> str:
     def editor_info(cls) -> dict[str, Any]:
         return {
             "editorInfo": {
-                "name": "Sublime Text",
+                "name": "vscode",
```
jfcherng, line 201: Fwiw, this seems to be a must.

jfcherng:
About handling `$/progress`, https://github.com/sublimelsp/LSP/issues/2494

jfcherng:
image

jfcherng:
It also keep auto filling chat question for me 🤔

TerminalFi:
1. That's copilot suggested follow up. Probably need different approach 😀 
2. Yea, will have to check how vscode does it

TerminalFi:
> image

I think these are both fixed in latest

TerminalFi:
I'm still working on the other two

1. Auto fill with suggested next question
2. What to do when asking while waiting

TerminalFi:
> I'm still working on the other two
> 
> 1. Auto fill with suggested next question
> 2. What to do when asking while waiting

1. Follow up is a link at the base of the chat panel
2. We now prevent user submitting a new question while awaiting the next question. This is inline with VS Code

TerminalFi:
The refactor is beautiful

PR #169, opened 2024-06-19
Implement `.copilotignore` support to ignore files

TerminalFi:


jfcherng:
As per https://github.com/TerminalFi/LSP-copilot/issues/98#issuecomment-1559154960, I think we should following `.copilotignore` rather than invent our own.

As per https://github.com/orgs/community/discussions/10305#discussioncomment-6069843, it seems that `.copilotignore` already works in VSCode.

LSP-copilot.sublime-settings
```diff
@@ -17,7 +17,10 @@
 		"local_checks": false,
 		"telemetry": false,
 		"proxy": "",
-		"completion_style": "popup"
+		"completion_style": "popup",
+		"copilot_ignore": [
+			"**/*.py"
```
rchl, line 22: `fnmatch` doesn't support globs

jfcherng: Fwiw, we can use https://github.com/facelessuser/wcmatch if we want. `wcmatch` is already on py38 deps for PC v4. But still, I think it's better to just follow `.copilotignore`.

timfjord:
It looks like this `.copilotignore` is basically a list of patterns to ignore in a format similar to the one introduced in the `copilot_ignore` setting.
So maybe we could rely on the `.copilotignore` by simply splitting it into lines and treating it like the `copilot_ignore` setting in the PR (we just need to think how to cache it properly)?

TerminalFi:
@jfcherng @timfjord I think we only need to cover somehow detecting new folders added to a window. And disabling text commands if file is ignored. 

But does the decorator approach work well here?

TerminalFi:
We could combine these decorators to maybe

```
def _must_be_active_view_not_ignored(*, failed_return: Any = None) -> Callable[[T_Callable], T_Callable]:
    def decorator(func: T_Callable) -> T_Callable:
        @wraps(func)
        def wrapped(self: Any, *arg, **kwargs) -> Any:
            if is_active_view(self.view) and not CopilotIgnore(self.view.window()).trigger(self.view):
                return func(self, *arg, **kwargs)
            return failed_return

        return cast(T_Callable, wrapped)

    return decorator
```

jfcherng:
> We could combine these decorators to maybe

It seems to make sense because `.copilotignore. works for all copilot features

timfjord:
Maybe to avoid dealing with edge cases, we could add a separate command to manually reload the `.copilotignore` file

TerminalFi:
@jfcherng I can't seem to run ci-fix ?

TerminalFi:
I was also thinking about updating statusbar `LSP-copilot (ignored)` if a file is ignored ? Thoughts?

jfcherng:
> @jfcherng I can't seem to run ci-fix ?
> 
> ```
> make ci-fix
> ========== fix: ruff (lint) ==========
> ruff check --fix .
> All checks passed!
> ========== fix: ruff (format) ==========
> ruff format .
> 14 files left unchanged
> ```

it runs and nothing needs to be changed.

jfcherng:
> Maybe to avoid dealing with edge cases, we could add a separate command to manually reload the `.copilotignore` file

I thought that we can have a ViewEventListener for view activation. When a view is activated (focused), read `.copilotignore` file and set `is_copilot_ignored` attribute to the view. Reading the file shouldn't takes to much time.

jfcherng:
> I was also thinking about updating statusbar `LSP-copilot (ignored)` if a file is ignored ? Thoughts?

Following up, we can just read `is_copilot_ignored` attribute of the activated view. `status_text` is now a jinja2 template. It should be quite flexible.

TerminalFi:
> > Maybe to avoid dealing with edge cases, we could add a separate command to manually reload the `.copilotignore` file
> 
> I thought that we can have a ViewEventListener for view activation. When a view is activated (focused), read `.copilotignore` file and set `is_copilot_ignored` attribute to the view. Reading the file shouldn't takes to much time.

I don't know that we need to read file per view. Probably need to update to ensure the following

1. File is read on new window
2. File is read when file is saved that name is `.copilotignore`
3. File is read when new folder added to window

Can 3. be achieved?

jfcherng:
> * File is read when file is saved that name is `.copilotignore`

There are potentially many ways to change that file. Not only ST's `save` command.

> File is read when new folder added to window

Why do we need this?

TerminalFi:
> > File is read when new folder added to window
> 
> Why do we need this?


A new folder in the `window.folders()` might have its own .copilotignore file

But if we just loaded them on view activated, then that wouldn't be a problem.

So it is 1. Load on vewi activated (Doesn't update if the file is modified outside of sublime still. or 2. file watch for .copilotignore

🤷

jfcherng:
> 1. Load on vewi activated (Doesn't update if the file is modified outside of sublime still.

Well, my original proposal is when a view gets activated

1. read `.copilotignore` file
2. check whether the current view should be ignored or not basing on the file read at step 1. 
3. set `is_copilot_ignored` attribute for the view

And then all other copilot functionality (and statusbar text) should check `is_copilot_ignored` of the current view.

TerminalFi:
I am happy with either proposal. I think there is file watch ? https://github.com/kapitanluffy/sublime-file-watcher

jfcherng:
> I am happy with either proposal. I think there is file watch ? [kapitanluffy/sublime-file-watcher](https://github.com/kapitanluffy/sublime-file-watcher)

It future depends on https://packagecontrol.io/packages/LSP-file-watcher-chokidar at https://github.com/kapitanluffy/sublime-file-watcher/blob/4177e3a9dd29537dd257a990f75ad2a2be9ba4a1/src/file_watcher_handler.py#L9-L10

I also created a PR for `watchdog` lib for py38 dep on https://github.com/packagecontrol/channel/pull/16.

jfcherng:
I just made a renaming https://github.com/TerminalFi/LSP-copilot/commit/8308bc02a8f68012d8e3c85e8feabebefdb399a1

jfcherng:
> File is read when new folder added to window

This seems to be an issue if we go the watcher way.

jfcherng:
Hmm.. users can drag sidebar folder to change to order of workspace folders. I guess this will break the current implementation?

TerminalFi:
> Hmm.. users can drag sidebar folder to change to order of workspace folders. I guess this will break the current implementation?

Why does order matter ? We store the folder prefix with its patterns. ?

TerminalFi:
> > File is read when new folder added to window
> 
> 
> 
> This seems to be an issue if we go the watcher way.

Yeah, we need ST support to fix all
Caveats

plugin/ui/completion.py
```diff
@@ -347,4 +346,7 @@ def hide(cls, view: sublime.View) -> None:
 
     @classmethod
     def close(cls, view: sublime.View) -> None:
-        del _view_to_phantom_set[view.id()]
+        try:
+            del _view_to_phantom_set[view.id()]
+        except KeyError:
+            pass
```
jfcherng, line 352: Can be simplified as `_view_to_phantom_set.pop(view.id(), None)`.

jfcherng:
I got the following error in console.
```
ValueError: ('is_enabled must return a bool', <LSP-copilot.plugin.commands.CopilotSignOutCommand object at 0x000002C0BB580EE0>)
```

TerminalFi:
> I got the following error in console.
> 
> ```
> ValueError: ('is_enabled must return a bool', <LSP-copilot.plugin.commands.CopilotSignOutCommand object at 0x000002C0BB580EE0>)
> ```

Yes, that is because of 

```
    @_must_be_active_view_not_ignored()
    @_provide_plugin_session(failed_return=False)
    def is_enabled(self, plugin: CopilotPlugin, session: Session) -> bool:  # type: ignore
        return self._can_meet_requirement(session)
```

The decorator can be used `@_must_be_active_view_not_ignored()` I don't think

plugin/__init__.py
```diff
@@ -1,7 +1,5 @@
 from __future__ import annotations
 
-from plugin.utils import CopilotIgnore
```
jfcherng: For pylance, there is

```
        // Defines the default format for import module.
        "python.analysis.importFormat": "relative",
```

PR #171, opened 2024-06-20
refactor: use Jinja2 for completion popup / panel completion

jfcherng:
This makes logics more clear.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,6 +14,9 @@ indent_style = tab
 [Makefile.*]
 indent_style = tab
 
+[*.jinja]
+indent_size = 2
+
 [*.md]
 indent_size = 2
 trim_trailing_whitespace = false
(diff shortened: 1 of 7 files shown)


timfjord:
@jfcherng I was going to extract the phantom stuff to the templates but wasn't sure about the template extension. 
You used the `.md` extension in the templates (e.g. `completion@popup.md.jinja`) but technically it is `.html`.
What if I rename `*.md.jinja` to `*.html.jinja`

jfcherng:
> `completion@popup.md.jinja`) but technically it is `.html`

No, it's Markdown. And at the same time, Markdown is a superset of HTML.

- https://github.com/TerminalFi/LSP-copilot/blob/dcbd65de368b6535ef3bc1cf86ea24fe58e0a2a7/plugin/ui/completion.py#L241
- https://github.com/TerminalFi/LSP-copilot/blob/dcbd65de368b6535ef3bc1cf86ea24fe58e0a2a7/plugin/templates/completion%40popup.md.jinja#L91-L93

Issue #117, opened 2023-08-07
popup width


cotcomsol:
Is there a way to change the popup width?  So far I have been unable to find a setting.  On my system it always shows up too narrow and the completions are cut off.  My sublime window in this screenshot is over 3000 pixels wide so plenty of space for it to display the full completion if it just made the popup wider.
![Screenshot_2023-08-07_09-06-29](https://github.com/TerminalFi/LSP-copilot/assets/414939/54ce4836-87f9-4685-a0ac-0199f54a5294)

TerminalFi:
The short answer, not currently. The long answer is it impossible believe to make this user a user configuration. But not sure if it is really a high priority. But I do see how this can make it not as usable.

timfjord:
Just a note, you could use
```json
"settings": {
  "completion_style": "popup"
}
```
to have a VSCode-like completion style.
But please note it is a bit buggy

TerminalFi:
After months, releasing this within the week

WangWei90:
> Just a note, you could use
> 
> ```json
> "settings": {
>   "completion_style": "phantom"
> }
> ```
> 
> to have a VSCode-like completion style. But please note it is a bit buggy

Can the "phantom" be setup with a different background color, can the "phantom" use the upright style?

timfjord:
@WangWei90 I don't think we set the background colour explicitly, but since the package follows [the best practices](https://www.sublimetext.com/docs/minihtml.html#best-practices) it should be possible to customize it if needed.

I'm not sure what the upright style means though. But if it is about styling, then it should be possible to customize it too.

WangWei90:
Thank you~ @timfjord 

I will try to test this later.

TerminalFi:
@cotcomsol Is this still desired?

cotcomsol:
It would definitely be helpful for me.  I find that most of the time the completion is cut off and so mostly useless.

jfcherng:
or maybe we show it when the time elapses a thrshold

TerminalFi:
> or maybe we show it when the time elapses a thrshold

Wrong thread :)

jfcherng:
Hey @TerminalFi are you going to review https://github.com/TerminalFi/LSP-copilot/pull/178 ? I see you close this issue without merging that PR.

jfcherng:
Hmm... I guess #178 won't help. The popup width is provided as an argument to a ST plugin API.

TerminalFi:
[...]

PR #176, opened 2024-07-04
Allow for passing custom CSS to modify popup

TerminalFi:
Potentially solves #117 But maybe we don't store in view settings and we rather just pass in session to allow grabbing settings

### DIFF
--- a/LSP-copilot.sublime-settings
+++ b/LSP-copilot.sublime-settings
@@ -10,6 +10,7 @@
 		"res"
 	],
 	"settings": {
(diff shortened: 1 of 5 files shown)


jfcherng:
Another way to do that is letting the user to create a CSS file and we load it in the Jinja template.

TerminalFi:
Yes, we could do that as well, I would think we need to make it very user friendly to know how to do that. 

Do you have any opinions?

jfcherng:
I think using file(s) are better. We can create menu for editing.

PR #178, opened 2024-07-07
feat: custom CSS for chat panel / popup completion / panel completion

jfcherng:
Potentially solves #117 and supersedes #176


jfcherng:
Maybe I shouldn't call it "override". They are actually "extra CSS rules".

TerminalFi:
One this to note, because we are caching, these changes to wont take affect till Plugin reload. So we either monitor for file changes or tell user to restart after editing

jfcherng:
> One this to note, because we are caching, these changes to wont take affect till Plugin reload. So we either monitor for file changes or tell user to restart after editing

Should be fixed with [`4e6c22a` (#178)](https://github.com/TerminalFi/LSP-copilot/pull/178/commits/4e6c22abf05b9fcf5382c36b2c67fa0e2ef971c8)

jfcherng:
I think this has been done now.

timfjord (approved):

TerminalFi:
Ship it

PR #194, opened 2024-07-29
feat: allow user customizable prompts for reusable logic

TerminalFi:
Copilot uses GPT in the background. Essentially meaning you can create prompts in the chat panel. This just allows users to add re-usable prompts that can be triggered by the same logic as the build in prompts

### DIFF
--- a/plugin/helpers.py
+++ b/plugin/helpers.py
@@ -214,12 +214,13 @@ def prepare_conversation_turn_request(
     for selection in view.sel():
         if selection.empty() or view.substr(selection).strip() == "":
             continue
+        file_path = view.file_name() or f"buffer:{view.buffer().id()}"
         selection_start = view.rowcol(selection.begin())
         selection_end = view.rowcol(selection.end())
         turn["references"].append({
             "type": "file",
             "status": "included",
-            "uri": filename_to_uri(view.file_name()),
+            "uri": file_path if file_path.startswith("buffer:") else file_path and filename_to_uri(file_path),
             "range": initial_doc["doc"]["position"],
             "visibleRange": {
                 "start": {"line": visible_start[0], "character": visible_start[1]},
--- a/plugin/helpers.py
+++ b/plugin/helpers.py
@@ -20,7 +20,7 @@ from .settings import get_plugin_setting_dotted
 from .types import (
     CopilotPayloadCompletion,
     CopilotPayloadPanelSolution,
-    CopilotRequestCoversationTurn,
+    CopilotRequestConversationTurn,
     CopilotUserDefinedPromptTemplates,
 )
 from .utils import (
@@ -197,10 +197,10 @@ def prepare_conversation_turn_request(
     message: str,
     view: sublime.View,
     source: Literal["panel", "inline"] = "panel",
-) -> CopilotRequestCoversationTurn | None:
+) -> CopilotRequestConversationTurn | None:
     if not (initial_doc := prepare_completion_request(view, max_selections=5)):
         return None
-    turn: CopilotRequestCoversationTurn = {
+    turn: CopilotRequestConversationTurn = {
         "conversationId": conversation_id,
         "message": message,
         "workDoneToken": f"copilot_chat://{window_id}",
--- a/plugin/helpers.py
+++ b/plugin/helpers.py
@@ -227,16 +227,17 @@ def preprocess_chat_message(
 
     if enum_has_value(CopilotConversationTemplates, message):
         is_template = True
-        message += " {{ sel[0] }}"
+        message += " {{ code }}"
     elif user_template := first_true(templates, pred=lambda t: f"/{t['id']}" == message):
         is_template = True
-        message = "{} {}".format(message, "\n".join(user_template["prompt"]))
+        message = "{} {}\n\n{}".format(message, "\n".join(user_template["prompt"]), "{{ code }}")
+
+    region = view.sel()[0]
+    lang = get_view_language_id(view, region.begin())
+    code = f"\n```{lang}\n{view.substr(region)}\n```\n"
 
     template = load_string_template(message)
-    lang = get_view_language_id(view, view.sel()[0].begin())
-    sel = [f"\n```{lang}\n{view.substr(region)}\n```\n" for region in view.sel()]
-
-    return is_template, template.render({"sel": sel})
+    return is_template, template.render({"code": code})
 
 
 def preprocess_completions(view: sublime.View, completions: list[CopilotPayloadCompletion]) -> None:
--- a/plugin/helpers.py
+++ b/plugin/helpers.py
@@ -224,20 +224,27 @@ def preprocess_chat_message(
     from .template import load_string_template
 
     is_template = False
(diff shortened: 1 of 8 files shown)


jfcherng:
I feel it's just weird to have `{{ sel[0] }}` in the prompt so I change it to [`17b7b1a` (#194)](https://github.com/TerminalFi/LSP-copilot/pull/194/commits/17b7b1a18e813400f5d93e25736b4f6117959551). But it also "always" appends the selected code to the end.

@TerminalFi What do you think?

TerminalFi:
> I feel it's just weird to have `{{ sel[0] }}` in the prompt so I change it to [`17b7b1a` (#194)](https://github.com/TerminalFi/LSP-copilot/pull/194/commits/17b7b1a18e813400f5d93e25736b4f6117959551). But it also "always" appends the selected code to the end.
> 
> 
> 
> At least I think that seeing `sel` as a list in the prompt is weird for non-plugin-dev users.
> 
> 
> 
> @TerminalFi What do you think?

Actually, I think that we are missing part of the payload. I need to test it. 

But I actually think it expects the code to be part of the references already. And then after the /review or prompt it expects either blank, or additional message from user. I'll post example here shortly

TerminalFi:
@jfcherng 

I haven't tested, but I think it might expect us to pass in a single reference

```
                    # References
                    # {
                    #     "type": "file", # Literal["file"]
                    #     "status": "" # Literal["included", "blocked", "notfound", "empty"]
                    #     "range": {} # CopilotPayloadCompletionRange,
                    #     "uri": ni.Type.String(),
                    #     "position": ni.Type.Optional(
                    #     "    ni".Type.Object({
                    #     "        line": ni.Type.Number({ minimum: 0 }),
                    #     "        character": ni.Type.Number({ minimum: 0 }),
                    #     "    "}),
                    #     ")",
                    #     "visibleRange": ni.Type.Optional(C2), # CopilotPayloadCompletionRange
                    #     "selection": ni.Type.Optional(C2), # CopilotPayloadCompletionRange
                    #     "openedAt": ni.Type.Optional(ni.Type.String()),
                    #     "activeAt": ni.Type.Optional(ni.Type.String()),
                    # }
```

jfcherng:
you mean the server will auto append codes if the template is built-in?

TerminalFi:
> you mean the server will auto append codes if the template is built-in so appending it again would cause duplication?

Yes, I think that when we send a command like 

`/review` If it contains no `code` snippet the service will then look at the first `ref` which contains the above details, such as visibleRange, Selection, and attempt to use those. And then if the user send `/review {{ code[0] }}` it will use the provided code instead of the ref.

Untested

jfcherng:
Ah yes, if there is no selected code for `/fix`, Copilot just say there is no code.

TerminalFi:
> Ah yes, if there is no selected code for `/fix`, Copilot just say there is no code. We can decide what is expected to happen however.
> 
> some cases to be dealt with:
> 
> * No caret
 
 Assume full file + visiblerange

> * 1 caret, no selected code

 Assume full file + visiblerange

> * 3 carets, only the 2nd has selected code (the 1st and 3rd has no selected code)

1st 3rd are dropped

jfcherng:
> > * 3 carets, only the 2nd has selected code (the 1st and 3rd has no selected code)
> 
> 1st 3rd are dropped

what if 2 carets, both have codes?

TerminalFi:
Also, I have a feeling, if we implement this, we can likely hack together `inline` which I believe Copilot then response with WorkspaceEdits.

But the above statement has a lot of assumptions

TerminalFi:
> Also, I have a feeling, if we implement this, we can likely hack together `inline` which I believe Copilot then response with WorkspaceEdits.
> 
> 
> 
> But the above statement has a lot of assumptions
> 
> 

Since they support multiple refs, let's test sending each non empty carets code snippet as refs

TerminalFi:
FWIW, JetBrains might show more details. It's also using this same node binary

TerminalFi:
> FWIW, JetBrains might show more details. It's also using this same node binary 

@jfcherng sre you working this? Have any local changes ?

jfcherng:
> > FWIW, JetBrains might show more details. It's also using this same node binary 
> 
> @jfcherng sre you working this? Have any local changes ?

no

plugin/helpers.py
```diff
@@ -214,12 +214,13 @@ def prepare_conversation_turn_request(
...
         if selection.empty() or view.substr(selection).strip() == "":
             continue
+        file_path = view.file_name() or f"buffer:{view.buffer().id()}"
         selection_start = view.rowcol(selection.begin())
         selection_end = view.rowcol(selection.end())
         turn["references"].append({
             "type": "file",
             "status": "included",
-            "uri": filename_to_uri(view.file_name()),
+            "uri": file_path if file_path.startswith("buffer:") else file_path and filename_to_uri(file_path),
```
jfcherng, line 223: is that `file_path and` required?

plugin/helpers.py
```diff
@@ -210,18 +257,33 @@ def preprocess_message_for_html(message: str) -> str:
...
-    if is_template:
-        message += " {{ sel[0] }}"
+    templates = templates or []
+    user_template = first_true(templates, pred=lambda t: f"/{t['id']}" == message)
+    is_template = False
+
+    if user_template:
+        is_template = True
+        message += "\n\n{{ user_prompt }}\n\n{{ code }}"
+        message += "\n\n{{ user_prompt }}\n\n"
```
jfcherng, line 274: weird duplication?

plugin/types.py
```diff
@@ -180,6 +198,25 @@ class CopilotPayloadConversationTemplate(TypedDict, total=True):
     scopes: list[str]
 
 
+class CopilotRequestCoversationTurn(TypedDict, total=True):
```
TerminalFi, line 201: Typo in `Coversation` -> `Conversartion`

plugin/types.py
```diff
@@ -180,6 +198,25 @@ class CopilotPayloadConversationTemplate(TypedDict, total=True):
...
+class CopilotRequestCoversationTurnReference(TypedDict, total=True):
+    type: str
+    status: str
+    uri: str
+    range: CopilotPositionType
+    visibleRange: CopilotRangeType
+    selection: CopilotRangeType
+
+
 class CopilotRequestCoversationAgent(TypedDict, total=True):
```
TerminalFi, line 220: Guess I made a lot of typos :D Typo in `Coversation` -> `Conversation`

TerminalFi (commented):
Few typos

TerminalFi:
Unrelated to this MR. I noticed all the Copilot Chat stuff stays in command palette when signed out

TerminalFi:
LGTM. We can refine in later MR's after feature lives for awhile

jfcherng (approved):

jfcherng:
supposedly these branches are no longer used?

-   fix/deleteTurn
-   fixup/Templates

TerminalFi:
> supposedly these branches are no longer used?
> 
> * fix/deleteTurn
> * fixup/Templates

deleted

PR #195, opened 2024-08-08
Impl/copilot toggle

TerminalFi:


plugin/ui/chat.py
```diff
@@ -117,8 +126,11 @@ def conversation(self, value: list[CopilotPayloadConversationEntry]) -> None:
     # normal methods #
     # -------------- #
 
-    def __init__(self, window: sublime.Window) -> None:
-        self.window = window
+    def __init__(self, window: sublime.Window | None) -> None:
+        if not window:
+            raise ValueError("window is required")
+
+        self.window: sublime.Window = window
```
jfcherng, line 133: What's going on? If we need this check, then the issue is probably in other places.

plugin/ui/chat.py
```diff
@@ -169,8 +181,8 @@ def close(self) -> None:
 
 class _ConversationEntry:
     def __init__(self, window: sublime.Window) -> None:
-        self.window = window
-        self.wcm = WindowConversationManager(window)
+        self.window: sublime.Window = window
+        self.wcm: WindowConversationManager = WindowConversationManager(window)
```
jfcherng, line 185: Tools should be able to auto recognize these types correctly?

plugin/utils.py
```diff
@@ -111,15 +111,25 @@ def fix_completion_syntax_highlight(view: sublime.View, point: int, code: str) -
     return code
 
 
-def get_copilot_setting(instance: sublime.Window | sublime.View, prefix: str, key: str, default: Any = None) -> Any:
+def get_copilot_setting(
+    instance: sublime.Window | sublime.View | None, prefix: str, key: str, default: Any = None
```
jfcherng, line 115: Allowing `None` smells bad imo...

plugin/ui/chat.py
TerminalFi: Ci-check was complaining

plugin/utils.py
TerminalFi: I can remove. Ci-check was just complaining

jfcherng:
I can't seem to reproduce those ci-check complaints so I just revised them back to what they were.

jfcherng (approved):

PR #201, opened 2024-08-16
feat: support copilot-1.39.0

TerminalFi:
### TODO

- [x] References mirror VSCode

VSCode references only consider the first cursor and considers all opened files. 

### Example

Let’s assume the three files are:

	1.	file1.py
	2.	file2.py
	3.	file3.py

And each file has the following first cursor/selection positions:

	•	file1.py: Cursor at line 10, character 5.
	•	file2.py: Cursor at line 20, character 15.
	•	file3.py: Cursor at line 30, character 25. Cursor at line 45, character 40
	
We will only reference

	•	file1.py: Cursor at line 10, character 5.
	•	file2.py: Cursor at line 20, character 15.
	•	file3.py: Cursor at line 30, character 25.
	
	
	
What needs to happen

```py
def prepare_conversation_turn_request(
    conversation_id: str,
    window_id: int,
    message: str,
    view: sublime.View,
    source: Literal["panel", "inline"] = "panel",
) -> CopilotRequestConversationTurn | None:
    if not (doc := prepare_completion_request_doc(view)):
        return None

    # References can technicaly be across multiple files
    # TODO: Support references across multiple files
    references: list[CopilotRequestConversationTurnReference | CopilotGitHubWebSearch] = []
    visible_range = st_region_to_lsp_range(view.visible_region(), view)
    for selection in view.sel():
        if selection.empty() or view.substr(selection).isspace():
            continue
        references.append({
            "type": "file",
            # included, blocked, notfound, empty
            "status": "included",
            "range": st_region_to_lsp_range(selection, view),
            "uri": filename_to_uri(file_path) if (file_path := view.file_name()) else f"buffer:{view.buffer().id()}",
            "visibleRange": visible_range,
            "selection": st_region_to_lsp_range(selection, view),
            "position": st_point_to_lsp_position(selection.begin(), view),
            # openedAt: ni.Type.Optional(ni.Type.String()),
            # activeAt: ni.Type.Optional(ni.Type.String()),
        })
```

This needs to be updated to follow the logic like 

```
for view in session(window_id):
    create_reference(view)
```

We need to list only the views for the window_id tied to the session

In effort to continue to show the value of this and maximize the usage of MiniHTML and drive more features we should expand to allow for expandable and collapsable references. :D


plugin/commands.py
```diff
@@ -568,6 +586,25 @@ def _on_result_conversation_agents(self, payload: list[CopilotRequestConversatio
...
+        session.send_request(Request(REQ_GET_PROMPT, {"doc": doc}), self._on_result_get_prompt)
+
+    def _on_result_get_prompt(self, payload) -> None:
+        window = self.view.window()
+        if not window:
+            return
+        view = window.create_output_panel(f"{COPILOT_OUTPUT_PANEL_PREFIX}.prompt_view", unlisted=True)
+
+        with mutable_view(view) as view:
+            view.set_syntax_file("Packages/JavaScript/JavaScript.sublime-syntax")
```
jfcherng, line 602: what's this for?

jfcherng:
I think we first have to update the those typedict schemas for the new server.

jfcherng:
[chore: bump server to 1.39.0](https://github.com/TerminalFi/LSP-copilot/pull/201/commits/8e58b62d15b58917073d0e49766642ae80c8aa22) for testing this PR

TerminalFi:
@jfcherng I think we might need to rethink how we store conversation long term. But it is almost working. However I am having issues with jinja2 templating. Mind taking a look?

The below code causes this error `Parse Error at line 351 column 28: closing p does not match opening div`
```
{%- if section.kind == "report" -%}
<div class="references">
{%- if section.references_expanded -%}
  <a class="reference_toggle" href='{{ section.toggle_references_url }}'>{{ section.references|length }} References</a>
  <div class="reference">
{% for reference in section.references %}
      <a href="#">{{ reference['uri'] }}</a>
{% endfor %}
  </div>
{% else %}
  <a class="reference_toggle" href='{{ section.toggle_references_url }}'>{{ section.references|length }} References</a>
{%- endif -%}
</div>
{%- endif -%}
```

jfcherng:
> The below code causes this error `Parse Error at line 351 column 28: closing p does not match opening div`

That's caused by mdpopup will handles mixed HTML+MD wrong iirc. I encountered that before.

jfcherng:
> I think we might need to rethink how we store conversation long term.

I see you inject `references_expanded` into the payload from Copilot. I think we should create a new type and keep the original payload type stay intact.

plugin/template.py
```diff
@@ -41,6 +42,10 @@ def _plugin_asset_path(asset_path: str) -> str:
     return f"Packages/{PACKAGE_NAME}/plugin/assets/{asset_path}"
 
 
+def sanitize_path(path: str) -> str:
+    return unquote(path.replace("file://", ""))
```
jfcherng, line 46: is this just https://github.com/sublimelsp/LSP/blob/fe795a145a0d7b666caf00d15abe6b22380168e9/plugin/core/url.py#L46 ?

Issue #225, opened 2024-12-21
Stopped working entirely


Gabriel-p:
A few days ago (maybe yesterday?) I noticed this plugin just stopped working entirely. Absolutely nothing works. I tried uninstalling and installing back, no changes.

I can Sign out (apparently), but there's no 'Sign in' command shown in the command palette.

These are the only messages I see in the console:

```
LSP: LSP-copilot: Supported execute commands: ['github.copilot.didAcceptCompletionItem', 'github.copilot.didAcceptPanelCompletionItem']
LSP-copilot: INFO: [certificates] Removed 1 expired certificates
LSP-copilot: INFO: [lsp] GitHub Copilot Language Server 1.236.0 initialized
LSP-copilot: WARNING: [lsp] Invalid telemetry setting: Expected object
LSP-copilot: WARNING: [lsp] Invalid  setting: Expected all values to match
LSP-copilot: WARNING: [lsp] Invalid telemetry setting: Expected object
LSP-copilot: WARNING: [lsp] Invalid  setting: Expected all values to match
```

It also blocks `LSP-pyright` so maybe there's something broken with LSP?

jfcherng:
I didn't notice anything weird at this moment.

jfcherng:
You may try to revert to https://github.com/TerminalFi/LSP-copilot/releases/tag/v0.3.2 to check whether it's caused by a recent update.

TerminalFi:
@jfcherng `telemetry` changes

```
telemetry: T.Optional(T.Object({
			telemetryLevel: T.Optional(T.String())
		}
```

`null` or `all`

TerminalFi:
@jfcherng We should probably add all skills to the ignoreSkills array unless we know how to supprot them.

```
'current-editor'

'project-labels'

'recent-files'

'git-metadata'

'problems-in-active-document'

```

We also should test these before release

TerminalFi:
> A few days ago (maybe yesterday?) I noticed this plugin just stopped working entirely. Absolutely nothing works. I tried uninstalling and installing back, no changes.
> 
> I can Sign out (apparently), but there's no 'Sign in' command shown in the command palette.
> 
> These are the only messages I see in the console:
> 
> ```
> LSP: LSP-copilot: Supported execute commands: ['github.copilot.didAcceptCompletionItem', 'github.copilot.didAcceptPanelCompletionItem']
> LSP-copilot: INFO: [certificates] Removed 1 expired certificates
[...]

PR #226, opened 2024-12-23
Update LSP-copilot.sublime-settings

TerminalFi:


jfcherng (approved):

jfcherng:
Weird. I am not affected and there is no change on https://github.com/github/copilot.vim side.

jfcherng:
I don't think people are willing to provide telemetry information voluntarily. We can just set `telemetry` to `null`.

TerminalFi:
Fixes #225

PR #260, opened 2026-02-05
Update to the latest server binary

jfcherng:
This PR...

- Refactors the download process of the server binary.
  - Before: Use npm to donwload. This relies on Node.js and also it will download server binaries of all supported platforms. The downloaded size is super large.
  - After: Directly download the server binary zip of the target platform from https://github.com/github/copilot-language-server-release/releases
- Use the latest server bin.
- Some changes to make client/server communication work for the latest server.

### DIFF
--- a/plugin/client.py
+++ b/plugin/client.py
@@ -249,11 +249,11 @@ class CopilotPlugin(AbstractPlugin):
 
     @staticmethod
     def get_version() -> str:
-        """Return this plugin's version. If it's not installed by Package Control, return `"unknown"`."""
+        """Returns this plugin's version. If it's not installed by Package Control, returns an empty string."""
         try:
             return json.loads(sublime.load_resource(f"Packages/{PACKAGE_NAME}/package-metadata.json"))["version"]
         except Exception:
-            return "unknown"
+            return ""
 
     @classmethod
     def get_account_status(cls) -> AccountStatus:


language-server/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "private": true,
     "dependencies": {
-        "copilot-node-server": "^1.41.0"
+        "@github/copilot-language-server": "^1.421.0"
```
rchl: I would just suggest a hardcoded tag in one of the files. Something like https://github.com/sublimelsp/LSP-rust-analyzer/blob/7baf7015930f54e26efa58fd3533371e75212dbe/plugin.py#L36-L36

Then renovate can be set up to read and update that tag. Example: https://github.com/sublimelsp/LSP-rust-analyzer/blob/1a8bdb9bc02b1304c84d43b1dc33352d59870541/renovate.json#L6-L19
(since this is a standard semantic tag, the `versioningTemplate` wouldn't be needed in that case)

plugin/client.py
```diff
@@ -142,15 +134,45 @@ def __init__(self, session: weakref.ref[Session]) -> None:
...
+            log_warning(f"Failed to download server: {e}")
+            return
+
+        rmtree_ex(cls.plugin_storage_dir(), ignore_errors=True)
+
+        decompress_buffer(
+            io.BytesIO(data),
+            filename=version_manager.THIS_TARBALL_NAME,
+            dst_dir=cls.versioned_server_dir(),
+        )
```
rchl: Binary needs to be made executable on Mac (and Linux I guess). Currently it's not.

renovate.json
```diff
@@ -2,5 +2,19 @@
...
+      "customType": "regex",
+      "managerFilePatterns": [
+        "/^plugin/constants\\.py$/"
+      ],
+      "matchStrings": [
+        "SERVER_VERSION = \"(?<currentValue>.+?)\""
+      ],
+      "datasourceTemplate": "github-releases",
+      "packageNameTemplate": "github/copilot-language-server-release",
+      "versioningTemplate": "regex:^(?<major>\\d+)-(?<minor>\\d+)-(?<patch>\\d+)$"
```
rchl: The default `versioningTemplate` will be fine in this case since it's a semantic version unlike for rust.

jfcherng:
> OK but I'm still not sure that people are not interested in seeing server or package version by default.

We can change/keep it later anyway. It's unrelated to the goal of this PR.

PR #269, opened 2026-02-26
refactor: use lsp_check_applicable command

rchl:
LSP had a `lsp_check_applicable` command for a while to handle copilot use case.

Minor type fixes in the process.

### DIFF
--- a/plugin/listeners.py
+++ b/plugin/listeners.py
@@ -2,15 +2,17 @@ from __future__ import annotations
 
 import re
 from collections.abc import Iterable
-from typing import Any
+from typing import Any, final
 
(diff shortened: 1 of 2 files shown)


jwortmann:
> @jwortmann I have set `enabled: false` globally for copilot but the `lsp_check_applicable` command actually makes it start still.

Yes, I didn't consider someone running the command while the config is disabled. We should probably add a check
```python
if not config.enabled:
    return
```
after these lines
https://github.com/sublimelsp/LSP/blob/ead1085da645c6ad71e2efa9a3104b276502e605/plugin/core/registry.py#L239-L242

PR #271, opened 2026-03-31
refactor: add ruff rules for sorting imports

rchl:
This is a semi-automatically created PR that adds ruff configuration for imports formatting.

~~It matches configuration included in the LSP package with the intention of consolidating import style across all LSP packages.~~

~~If you have strong preference to a different style, please keep the configuration but update it to your liking so that anyone working on the package does not have to guess the correct style to use. See [ruff isort settings](https://docs.astral.sh/ruff/settings/#lintisort).~~

Rules are adapted to match the repo style.

(Since this PR was created semi-automatically, it might require some changes before being considered ready.)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -87,10 +87,8 @@ select = [
   "UP",
   "FURB",
   "SIM",
-  "F401",
 ]
 ignore = ["E203", "SIM102", "SIM108"]
-preview = true
 
 [tool.ruff.lint.isort]
 case-sensitive = false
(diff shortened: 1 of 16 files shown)


pyproject.toml
```diff
@@ -87,5 +87,15 @@ select = [
   "UP",
   "FURB",
   "SIM",
+  "F401",
 ]
 ignore = ["E203", "SIM102", "SIM108"]
+preview = true
```
rchl:
```suggestion
]
ignore = ["E203", "SIM102", "SIM108"]
```

predragnikolic: should we use `preview = true` as in other repos?

rchl: It's already there in the ruff section.

rchl:
Should not be controversial since it matches project's style.

PR #283, opened 2026-05-22
refactor: mgrating from AbstractPlugin to LspPlugin

jfcherng:
https://lsp.sublimetext.io/migrating_to_lsp_plugin/


plugin/client.py
```diff
@@ -243,16 +198,14 @@ def parse_proxy(proxy: str) -> NetworkProxy | None:
...
-        return None
-
-    @override
-    def on_settings_changed(self, settings: DottedDict) -> None:
-        super().on_settings_changed(settings)
+        context.configuration.init_options.update(editor_info)
+        context.variables["server_path"] = str(version_manager.server_path)
 
+    def on_initialized_async(self) -> None:
+        """Called once after the `initialized` notification, replaces on_settings_changed."""
```
rchl: I would think that it is not relevant to document what legacy api it replaces.

plugin/client.py
```diff
@@ -243,16 +198,14 @@ def parse_proxy(proxy: str) -> NetworkProxy | None:
...
+        if network_proxy := parse_proxy(context.configuration.settings.get("proxy") or ""):
             editor_info["networkProxy"] = network_proxy
 
-        configuration.init_options.update(editor_info)
-        return None
-
-    @override
-    def on_settings_changed(self, settings: DottedDict) -> None:
-        super().on_settings_changed(settings)
+        context.configuration.init_options.update(editor_info)
```
rchl:
```suggestion
        context.configuration.initialization_options.update(editor_info)
```

plugin/client.py
```diff
@@ -354,18 +307,20 @@ def update_status_bar_text(self, extra_variables: dict[str, Any] | None = None)
                 log_warning(f'Invalid "status_text" template: {e}')
         session.set_config_status_async(rendered_text)
 
-    def on_server_notification_async(self, notification: Notification) -> None:
-        if notification.method == "$/progress":
-            token = notification.params.get("token", "")
-            params = notification.params.get("value")
+    def on_server_notification_async(self, notification: ServerNotification) -> None:
```
rchl:
```suggestion
    @override
    def on_server_notification_async(self, notification: ServerNotification) -> None:
```

plugin/client.py
```diff
@@ -243,16 +198,14 @@ def parse_proxy(proxy: str) -> NetworkProxy | None:
...
-        configuration.init_options.update(editor_info)
-        return None
-
-    @override
-    def on_settings_changed(self, settings: DottedDict) -> None:
-        super().on_settings_changed(settings)
+        context.configuration.init_options.update(editor_info)
+        context.variables["server_path"] = str(version_manager.server_path)
 
+    def on_initialized_async(self) -> None:
```
rchl, line 208:
```suggestion
    @override
    def on_initialized_async(self) -> None:
```

rchl (commented):
Some notes:
- `VersionManager` could get the package storage path from `CopilotPlugin.plugin_storage_path`. Just to have a single source of truth for the path.
- TextCommands don't need SESSION_NAME to be set anymore
- `get_plugin_setting` is unused
- `get_plugin_setting_dotted` is only used to read the debug setting which feels a bit overkill to me. Especially as it's the only reason that `jmespath` is required. Also note that the `debug` setting is not a server settings so it would be better to have it in the root of the configuration file rather than in `settings.*`. One reason is that the server doesn't care about this setting and another is that changing settings in the root restarts the server while changing `settings` doesn't.

plugin/client.py
rchl, line 311:
```suggestion
    @override
    def on_server_notification_async(self, notification: ServerNotification) -> None:
```

jfcherng:
> get_plugin_setting_dotted is only used to read the debug setting which feels a bit overkill to me. Especially as it's the only reason that jmespath is required. Also note that the debug setting is not a server settings so it would be better to have it in the root of the configuration file rather than in settings.*. One reason is that the server doesn't care about this setting and another is that changing settings in the root restarts the server while changing settings doesn't.

For historical reason, almost all settings in `settings` are not server settings but plugin settings 😢

PR #305, opened 2026-07-17
fix(models): skip non-policy-gated models in model picker

jfcherng:
copilot/setModelPolicy always fails for models without a modelPolicy field in copilot/models (e.g. auto, completion-only models), so selecting them from the quick panel errored with "Failed to accept model policy". Filter them out before showing the picker.


jfcherng:
hopefully fixes https://github.com/sublimelsp/LSP-copilot/issues/302#issuecomment-5000179046

